PR #7, opened 2020-04-13
Add GitHub actions

Bara:
I also removed all sourcemod include files, it doesn't make sense for me to provide this.


Bara:
Looks good now: https://github.com/Bara/Surftimer-olokos-public/runs/585396119

PR #9, opened 2020-04-14
added onejumplimit remover to mapsettings

13ace37:
> Samatazz: should add a command to allow all zones more than one jump instead of manually setting them in each zone setting
>  Samatazz: i didnt mean a command to disable the anti bhopping for every map globally, just a command to update it for specific maps where you need to bhop more than once e.g surf_omnific


olokos:
So happy to see you sending pull requests! :)

PR #10, opened 2020-04-14
Speed missing HUD hotfix - Update misc.sp

sashajov:
Enabling the speed when passing the CP / Stage...

https://github.com/olokos/Surftimer-olokos-public/issues/5

I noticed there is a problem with the lower limit of the allowed characters in the center hud after this famous Valve's update. I fixed it by compressing color hex codes in misc.sp file.

I know that's not fully solving the problem, that's just a fix for enabling speed, but I am sure that olokos developer will create a better fix than me.

Thanks. :)


addons/sourcemod/scripting/surftimer/misc.sp
```diff
@@ -1909,7 +1909,7 @@ stock void PrintChatBonus (int client, int zGroup, int rank = 0)
 
 			RecordDiff = g_fOldBonusRecordTime[zGroup] - g_fFinalTime[client];
 			FormatTimeFloat(client, RecordDiff, 3, szRecordDiff, 54);
-			Format(szRecordDiff, 54, "-%s", szRecordDiff);
+			Format(szRecordDiff, 54, "%s", szRecordDiff);
```
olokos, line 1912: are you sure this is safe?

sashajov:
I'm not sure, but I think it is. :) I tried to shorten the codes and delete u / s, but it wasn't enough. Two characters were missing just enough for speed to appear. I decided to delete - / + in SR: and PB:

I've been using this on my servers for some time now. I hope you find what's safe and what's not.

sashajov:
> Looks good, maybe not the most elegant solution, but makes perfect sense.

Thanks, that's just an idea. :)

> Question is though, are you sure those -% and +% can be safely removed, like you did in this pull request?
> I have commented on the first one so you can see what I am exactly asking about :)

Honestly, I'm not sure, but I think it is. :) I tried to shorten the codes and delete u / s, but it wasn't enough. Two characters were missing just enough for speed to appear. I decided to delete - / + in SR: and PB:

I've been using this on my servers for some time now. I hope you find what's safe and what's not.

samatazz:
the fix i implemented was to just use short codes for the colours, saved me enough space to be fine, i also removed all spaces between the /'s to save space there too and i haven't had any issues arise yet.

the +'s and -'s are to allow the user to see when they're ahead or behind times, removing them can be ok due to the colour coding but otherwise a bit weird

olokos:
Very good idea, I appreciate it!
I went to the bottom of this and like @samatazz said + and - is just adding it in front of the string hiding behind %s, so there's no possibility of memory leak.

When I get a little bit more time I'll test all the options and implement this trying to lose as little code as possible.

Shortening the color codes is awesome and we'll definitely implement it, I'd rather have u/s in there if possible.
Isn't my point of view exactly what you implemented @samatazz ?
If yes, could you take a screenshot of how it looks like? That would save me some time :)

sashajov:
> When I get a little bit more time I'll test all the options and implement this trying to lose as little code as possible.

I am sure you will find the best solution...

> Shortening the color codes is awesome and we'll definitely implement it, I'd rather have u/s in there if possible.

I am very glad that my idea is useful. :)

I tried to remove spaces instead of u/s and +- but the formatting was screwed. If it is possible to save would be awesome! Missing u/s and -+ is annoying me s lot xD

> Isn't my point of view exactly what you implemented @samatazz ?
> If yes, could you take a screenshot of how it looks like? That would save me some time :)

This is from my server: https://i.imgur.com/9gJIBr7.jpg

olokos:
I think you are using Fixcolorhint by francisco, right?
This plugin is completely not needed as @13ace37 has fixed this bug within the code of surftimer already!
Possibly removing that plugin would allow more characters, can you try it out?
I'm currently stumped by University projects, they doubled the amount cos of coronavirus, that's why I am not testing it myself instantly :P

sashajov:
> I think you are using Fixcolorhint by francisco, right?
> This plugin is completely not needed as @13ace37 has fixed this bug within the code of surftimer already!

True! I forgot to turn off :)

> Possibly removing that plugin would allow more characters, can you try it out?

I will now open the test server and fully test it.

> I'm currently stumped by University projects, they doubled the amount cos of coronavirus, that's why I am not testing it myself instantly :P

Yeah, I understand :) I hope you will handle it asap.

olokos:
Sorry @sashajov
I've accepted #37 and can't now accept your pull request due to conflicts. 

Could you please just have the replacing of hint color text to csgotext in pull request?

Not sure if the codes that @samatazz used are correct, could you compare original colors and see whether your hex codes are more on point than the ones he used?

If you select a closer which is closer to samatazz, please include both csgohudtext and your color codes to the updated pull request, I'd love to merge it. 

Sorry for the hassle, it's my bad.

olokos:
#37 
Already introduces those fixes and doesn't remove + and - from the list.
Thank you for your effort though! <3

Issue #5, opened 2020-04-13
[BUG] Speed is missing from hud when entering a new stage. 


aKjeller:
**Describe the bug**
When the hud is displaying time difference the speed goes blank. 

**Screenshots**
![bild](https://user-images.githubusercontent.com/44802724/79114516-c5be6b00-7d83-11ea-8120-6121d4b26527.png)

**Plugin List**
SurfTimer and z4lab mapchooser

**Additional context**
(An improvment would be if it displayed the prestrafe speed leaving the stage zone.

sashajov:
I noticed there is a problem with the lower limit of the allowed characters in the center hud after this famous Valve's update. I fixed it by compressing color hex codes in misc.sp file.

E.g. #FFFFCC is same as #FFC

You can find short but similar or even same color hex code by using this web script: http://maettig.com/?page=PHP/CSS_Color_Converter

It's not enough because I was also forced to remove the u/s after the speed and also + and - before SR and PB to make enough space.

This is just fix for enabling speed, but I hope that developers will fix it aswell.

I will be happy if olokos fix it in his build because then I will not need to change these things each time.

olokos:
Can anybody check if the pull request that I have mentioned above by covert has resolved this issue?
In that case we wouldn't have to make color codes any shorter.

olokos:
Nope, #13 is irrelevant to this issue.
We've got to get rid of the text, please try the latest code and see how much we can actually keep, I'm crashing to bed now. 👼

CustomStats:
#37 Did not seem fix the issue of speed disappearing when going through stages. This also seems to have brought back the issue where the spacing on the HUD messes up, but this was already noted in the PR. 
https://streamable.com/8ba2eo

olokos:
I agree, the problem is still there, we should try some different approach.
Maybe we can try removing -/+ or at least u/s to just u.
Currently 25 hours since I last slept, so going to bed. :)

qawery-just-sad:
u think that if we reaise the value of `#define MAX_HINT_SIZE 225` to something like 255 or 512 that might help?

qawery-just-sad:
here is a look at `#define MAX_HIN_SIZE 255` [video](https://youtu.be/6c0LliXYWaA) going beyond 255 makes no difference

sashajov:
@qawery-just-sad  Where it is defined, I can try right now.

breakthegreat:
[...]

PR #13, opened 2020-04-15
fix memory leak

covertxd:
https://github.com/Franc1sco/FixHintColorMessages/issues/9


olokos:
This gives me vibes that it could potentially help with #5

PR #18, opened 2020-04-16
Fix SQL Hostname Error

sneak-it:
This solves an SQL error when your server name has symbols or quote marks in it by properly escaping the value of `hostname`.


olokos:
As long as we're sure that szEscServerName is actually assigned a value it looks great to me.
Could potentially help get memory usage down apart from the fix you've mentioned in commit message.

PR #20, opened 2020-04-16
Set bots' movetype to noclip

deadw1nter:
So that they won't fly around like a headless fly and look smooth

### DIFF
--- a/addons/sourcemod/scripting/surftimer/replay.sp
+++ b/addons/sourcemod/scripting/surftimer/replay.sp
@@ -696,8 +696,12 @@ public void LoadRecordReplay()
 
 		// "Having a bot in noclip and zero gravity ensures it's smooth" - Crashfort
 		// https://github.com/crashfort/SourceToolAssist/blob/be9218583ee0a8086c817a5bd29101b2a260e5a7/Source/surf_segmentplay.sp#L113
+		// Disabling noclip, makes the bot bug.
+		// Since the bot is getting teleported with angle and speed not angle and position
+		// the bot sometimes will fly around like a headless fly (because of getting blocked) or choppy on a ramp
 		SetEntityMoveType(g_RecordBot, MOVETYPE_NOCLIP);
-
+		SetEntityGravity(g_RecordBot, 0.0);
+		
 		PlayRecord(g_RecordBot, 0, 0);
 		// We can start multiple bots but first we need to get if bot has finished playing???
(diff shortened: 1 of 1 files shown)


deadw1nter:
> Not sure why did you remove the comment, I think it's best to keep it, for future reference.

The reason why I removed SetEntityGravity is it is not needed. Since replay-bot is [ getting teleported with angle and speed](https://github.com/olokos/Surftimer-olokos/blob/dev/addons/sourcemod/scripting/surftimer/replay.sp#L1157) not angle and position, replay-bot sometimes will fly around like a headless fly (because of getting blocked) or choppy on a ramp. To address this issue, we just need to set its movetype to noclip so that it will look smoother.

> Not really crucial to code itself, but can possibly help quite a bit in the future.

I didn't think about that, sorry :P

> Have you tested this change?

No, but I know without setting its movetype to noclip, replay-bot will look choppy (although it happens sometimes).

olokos:
Tbh when bots weren't crashing they rarely had the issue you're mentioning, that's why I'm saying we've got to test this commit before implementing.

There's quite a bit of changes coming in and with the level of complexity let's make sure this doesn't cause any issues, as soon as you or somebody confirms it causes no issues, I'm all in for merging it now.

deadw1nter:
Closed since it is buggier when noclip is enabled :((((((((((((((((((((((((

covertxd:
Would it make any difference in doing this?

```
SetEntProp(g_RecordBot, Prop_Data, "m_CollisionGroup", 2);
SetEntityMoveType(g_RecordBot, MOVETYPE_NOCLIP);
```

13ace37:
You should make sure to set 
```cpp
g_iDisableTriggers[client] = false;
````
for the bots, otherwise, they will ignore trigger_teleport.

olokos:
So have we come to the final conclusion on this?
Should we keep the bots noclipping or not?
I've reopened the issue, I think we need to decide together on the approach here.

13ace37:
Is this working?

deadw1nter:
@13ace37 As far as I try, it works as I expect. But when the bot trigger a trigger_push, it will "freeze" for a moment then turn back to normal. I tried to ignore trigger_push for bots but didn't work.

olokos:
Could you update to the latest 284 relase and see if it is any better from the noclipping bots?
I'm wondering if current state is better or if this should be merged at this point.

deadw1nter:
Closed since I don't have any time to test it, sorry.:(

olokos:
@ghost
You can reopen this PR, somebody eventually will come and test it, there's nothing bad with open PR if it's potentially good and might be merged. 😄

PR #21, opened 2020-04-16
Repo Cleanup

sneak-it:
The purpose of this PR is to clean up the repo.

I saw in 6c432295033c327177e901e61622253714c7ff16 that SourceMod includes were re-added after being removed in #7 by @Bara , which is bad practice. Includes could change at any time and differ between SM versions. The timer should be compiled with the correct includes for the major version it is compiled against.

That being said, here is what was changed:
- Removed all includes except surftimer include
- Removed all gamedata (all unnecessary)
- Removed hidden cvar check
- Removed all third party dependencies (included dhooks would not even function on current SourceMod stable build)
- Removed non-include files included with base SourceMod (geoip database, languages config, why were these even in there?)
- Replaced included Steam ID's that would grant certain users titles with non-player Steam ID's
- Updated readme heavily: reworded some things to be a bit more clear, removed bit about the 'only open source surf timer that is updated' (untrue, see influx), included myself to contributors (I contributed to the original ck years back and to several forks since then), included links to all required and recommended extensions, updated installation and compiling instructions, etc


olokos:
Good stuff going on! Thanks!
I have some buts though. 
By keeping the includes, extensions and not offloading them for user to download manually we allow for the ease of use for newcomers to surf server hosting. 

Additionally by keeping the includes building the surftimer is as simple as dropping compile.exe and spcomp.exe, double click and that's it. 

That's my personal workflow, but if you've got a better one, please do let me know. 

I can be in charge of updating the includes when necessary, it's not really that much hassle. 

By removing all of those dependencies we're allowing more room for an error for less tech-savvy users and bug reports on the same code can result in different errors, potentially. 

I would be for keeping the dependencies, we could update them as you've spotted outdated dhooks for example. 

In the long run this can potentially cause less issues and allow for a less steep entry for fresh server owners. 
Good point about influx, I've thought we're the only ones. 

Maxmind has now deprecated geoip1 db and it's not possible to download it anymore from their website, so either we rewrite code to support the new geoip2 format used by Maxmind or we should keep it in the repo. 

Otherwise vast majority of sever owners using our repo won't have any geoip functionality.

sneak-it:
Thanks for being so active!

I'm all for catering to newcomers, and I think that #7 is there for them - if anyone is that new, they shouldn't even have to compile themselves. Pre-compiled releases are the ideal way to go, just like most major projects here on github. I strongly believe following basic installation instructions should be required if someone **chooses** to compile themselves. For everyone actively contributing, we should poll and see if this is the best way to go - but based upon the previous PR from Bara and what I've heard from others, this is the preferred way to move forward.

The latest known legacy GeoIP database is already included with every download of SM, so we don't need to include it here.

olokos:
It's my pleasure, for so long I've been trying to push this cart myself and now the cart got wings and is going hyperspace, this makes me so happy!

AFAIK currently the github actions have no release artifact published, but it's very easy to change from what I've read briefly.

I'd say you've convinced me entirely about this. 

@covertxd also thinks this the same way you do and so did Bara, so at this point I'll leave it for you to decide if we are going to make a strawpoll and decide as a group or just merge this already. 

I've got to say I haven't thought about possible licensing issues.

If you are just going to merge this, please do add buld artifacts to github actions while at it, so newbies would be in a safe spot.

Bara:
If you want Artifact uploads, I recommend to reduce the size of every build package. As example by removing extensions and other not really needed files.

Bara:
@sneak-it and please use in the future the `dev`-branch as base branch. master branch will be used as "release"-branch. 😉

LubricantJam:
I agree with what sneaK has said here, this is the best standard to follow and the workflow pretty much builds itself for users to use. Maybe to make it even easier, it creates a release for them.

sneak-it:
@Bara - this would break the GH action, no?

Bara:
Yes would break it, but should easy adn quick to fix. 🤔

PR #22, opened 2020-04-16
Added Cvar to disallow !pause

Moisty145:


13ace37:
why? there is ck_pause ?

LubricantJam:
I'll explain why this is, it's due to some people running blacky's movement anticheat the !pause due to the MoveType_None that it uses. It's typically very situational, hence the cvar.

13ace37:
Just remove the whole command. It's not working correctly anyway.

LubricantJam:
Didn't know if people wanted it to be kept, hence why we added the cvar for it. If people want it removed, we can do so.

13ace37:
But I can't see any changes to 
```cpp
	if (GetConVarBool(g_hPauseServerside) == false && client != g_RecordBot && client != g_BonusBot)
		{
			CPrintToChat(client, "%t", "Pause1", g_szChatPrefix);
			return;
		}
```

It's also returning, why would you add a second cvar to do the same thing?

sneak-it:
> I'll explain why this is, it's due to some people running blacky's movement anticheat the !pause due to the MoveType_None that it uses. It's typically very situational, hence the cvar.

I have a similar function in my AC - I will PR in a new native that returns if a client is paused. You will then be able to use this native in the AC to ignore players that are paused.

Moisty145:
Didin't realize there was a lready a cvar for this. Why is it placed so late though? surely if you disallow pausing you might as well stop it at the start?

olokos:
I'm not sure what's the final decision about mering cvar to allow/disallow pause is, but it looks good to me.
Second commit is definitely useful, so I'm merging this.

PR #26, opened 2020-04-17
Update README.md - include trails chroma

olokos:
Currently Trails Chroma are being used in the code, so we should also mention that as a recommended and also remind people to put materials on fastdl.

I created this PR as I am not sure if this should be in required or recommended or completely replaced in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Yes, but not any more than other projects currently and for past years.
 
 **Recommended:**
 * [Cleaner Extension](https://github.com/Accelerator74/Cleaner) (Suppresses console warnings)
+* [Trails Chroma](https://github.com/Nickelony/Trails-Chroma) (Pretty trails, also recommended to have the materials on fastdl)
 
 ## Issue Rules
 


sneak-it:
The code from that Trails plugin is integrated into the timer itself, I have no idea why. The materials themselves are hosted here in the repo already (materials/trails). Nickelony is already credited at the bottom, no further action or dependencies are needed from it.

I think we should remove this integration, as far as I can tell there is nothing in this code that warrants it being a permanent part of the timer. If you want to use it separately and offer it to your VIP's, you can do so with overrides. If there was an issue with hiding trails, we should be able to do so with some proper natives.

@13ace37 What do you know about this?

13ace37:
I added the trails to give the VIPs of z4lab access to the trails. Since we changed to the "custom title system". Also, I'm not really sure what it does anymore, but there is a cvar called `ck_trails_enabled`. It should stop the timer from loading the trail files, so server owners can use it without putting any trail related files on the server. (But I'm not 100% sure!).

olokos:
It's already included in readme and trails chroma are gone from surftimer's code.

PR #29, opened 2020-04-17
Update README.md - we dont support SM1.11

olokos:
I created this pull request mainly to get your input on the Repository description itself.
I've heard words that it's not ideal and I agree, but what do you propose to set the repository description to?
Also any improvements to this readme updates are welcome!


sneak-it:
I would just keep it simple and say we only support the latest stable version of Sourcemod.

olokos:
Looks good now? :P

PR #32, opened 2020-04-17
Bug Fix and DB Query optimization for profile loading

ashakiri-dev:
This PR has two fixes from my personal fluffyst fork

1. Fix to !startpos on staged maps
The startpos command has a bug where, on a staged map, if you have set a custom start position, the commands to switch stages (e.g. !s, !back) will not bring you to the stage - it will just teleport you to the custom startpos. This bug is fixed

2. Significant speed-up for single user profile load
Profile load tends to be very slow on maps with a large stage/cp count with several clears. This is due, in part, to poor query design. Ranks, for example, are calculated by selecting all clear times faster than the player, sorting them, then returning that to the plugin. The plugin then does a GetRowCount'. This can be significantly improved by simply using SELECT COUNT(*) instead. This commit implements this for profile load, and is intended to serve as a proof of concept for implementing throughout the plugin.

Before: 

After:


olokos:
Comparison of query times on a huge DB
Courtesy of @ashakiri-dev

PR #33, opened 2020-04-17
Add index file, fix for style variables not resetting.

dPexxIAM:
adds index to ck_wrcps table to help with one of the worst-performing/most called queries

Fix for style variables not getting reset and new stage records being shown as (2/1) instead of (1/1)

Fixed queries for stage/bonus count on !profiles


olokos:
Have you tested those changes after recent updates to dev branch?

dPexxIAM:
Index file has been tested but the queries I changed weren't tested. I assumed the issues were still present. I can test if need be 

https://discordapp.com/channels/366959507764674560/375452719529328644/700210903936794683

olokos:
Please test both the queries and index combo, so we could merge it with dev for testing on a bigger scale.

dPexxIAM:
The queries don't affect anything, it's just a work around for the styles variables not getting reset:

Styles variables don't get reset on mapchange, so if the next map has no style records, it will compare to prev map
Sometimes it shows rank 2/1, this happens because runtime is a float value and not decimal, and since the query compares times it's not accurate. It can be fixed by substracting a epsilon to the runtime part, kinda like this: 

SELECT count(runtimepro)+1 FROM `ck_playertimes` WHERE `mapname` = '%s' AND `runtimepro` < %f-(1E-3);

dPexxIAM:
When I have time I can test the changes myself

sneak-it:
Why not just reset the vars on map change instead of adding complications to the query?

dPexxIAM:
most of the style related variables are dupes which make maintaining the code a pain,  If style variables get refactored at some point the good thing to do would be adding a new index to all the duplicate variables. this fix suffices for now

PR #38, opened 2020-04-18
remove trails chroma from timer

13ace37:
https://github.com/olokos/Surftimer-olokos/issues/36#issuecomment-615484862


Bara:
Would make both files obsolete, so you can remove them too

sneak-it:
As long as it compiles without any warnings/errors, looks great! 👍

olokos:
Seems like there are some conflicts, quite possibly because of the recent merges of pull requests i just did.

Could you confirm it works fine and causes no issues with current code? 
@13ace37

Bara:
I can't see the conflicts, but should comes from the enumStrucs merge because I modified the enum array.

olokos:
Yeah, exactly what I thought, but I'm on mobile, so can't resolve the conflict in this PR myself, so I can't merge it myself, until conflicts are resolved.

13ace37:
Never compiled it btw. I think there is still a cvar left.

PR #43, opened 2020-04-20
Fix prespeed exploit

### LINKED ISSUE #42: [BUG] Prehopping over speedcap without prehop limit
**Describe the bug**
You can exceed the start zone speed cap if you leave the zone and instantly jump back it.

**Expected behavior**
Speed limit in the startzones even if the user left it.

**Screenshots**
https://youtu.be/K75gcq7ej7Y

**Plugin List**
olokos@dev 

**Additional context**
N/A

sneak-it:
Fixes #42 

The bhop detection logic is very sub-par and should be re-worked. This should be thoroughly tested, even though I checked to make sure that this patched the exploit.


sashajov:
Hi, any news about prespeed glitches?

PR #51, opened 2020-04-23
Further optimizations to sql.sp

justaFluffypanda:
After reading [ashakiri's](https://github.com/olokos/Surftimer-olokos/commit/6de0d5bab72a898e7bf5a11362e1f4f4bd8c482c) post detailing the poor performance of using the GetRowCount() function I have gone through and replaced it with faster FetchInt() and FetchFloat() functions, along with modifying the original queries to support the use of FetchInt() and FetchFloat().

I have also heavily modified the SQL_db_CalcAvgRunTimeCallback() function which, for some reason, was loading the entire runtimepro column (12000+ entries on one of my servers alone) to derive the average from the column using a while loop. I've replaced this with a new query in queries.sp that simply returns the average from the column and have modified sql.sp to work with this value, if there was some method to the while loop madness I couldn't see it.

The rest of the changes in the 2nd commit are just indentation changes to remove all the compiler warnings.  I have had his build live on 3 of my servers for ~24 hours, with no apparent problems yet.

### DIFF
--- a/addons/sourcemod/scripting/surftimer/sql.sp
+++ b/addons/sourcemod/scripting/surftimer/sql.sp
@@ -1450,6 +1450,7 @@ public void db_viewPlayerPointsCallback(Handle owner, Handle hndl, const char[] 
 	if (SQL_HasResultSet(hndl))
 	{
 		int style;
+        
 		while (SQL_FetchRow(hndl))
 		{
 			style = SQL_FetchInt(hndl, 10);
@@ -1463,7 +1464,7 @@ public void db_viewPlayerPointsCallback(Handle owner, Handle hndl, const char[] 
 				g_iTotalConnections[client] = SQL_FetchInt(hndl, 8);
 			}
 		}
-
+        
 		g_iTotalConnections[client]++;
 
 		char updateConnections[1024];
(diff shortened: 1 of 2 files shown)


addons/sourcemod/scripting/surftimer/sql.sp
```diff
@@ -9078,7 +9072,7 @@ public void db_viewCustomTitles(int client, char[] szSteamID)
 	Handle pack = CreateDataPack();
 	WritePackCell(pack, client);
 	WritePackString(pack, szSteamID);
-	Format(szQuery, 728, "SELECT `title`, `namecolour`, `textcolour`, `inuse`, `vip`, `zoner`, `joinmsg` FROM `ck_vipadmins` WHERE `steamid` = '%s'", szSteamID);
+	Format(szQuery, 728, "SELECT `title`, `namecolour`, `textcolour`, `inuse`, `vip`, `zoner`, `joinmsg` FROM `ck_vipadmins` WHERE `steamid` = '%s';", szSteamID);
```
olokos, line 9075: are you sure about this semicolon? 😉

olokos:
Nevermind, the semicolon is where it should belong!

addons/sourcemod/scripting/surftimer/sql.sp
justaFluffypanda: I do believe it needs to be there, yes. I have not explicitly tested this exact functionality, but all  the format() function calls within sql.sp either specify an existing query initialized in queries.sp or they provide an inline query inside double quotes, as is the case here. As such it should need the terminating semicolon.

olokos:
@justaFluffypanda 
Sadly this causes
`surftimer/hooks.sp(1370) : error 035: argument type mismatch (argument 1)`
During compilation

justaFluffypanda:
Mine compiled just fine, I'll check it later today.

PR #61, opened 2020-04-25
this would resolve #39 by deleting silent spec

### LINKED ISSUE #39: [BUG] Silent spec not working correctly
**Describe the bug**
When spectating with silent spec enabled only other spectators don't see the one with silent spec but the person that is spectated sees everyone.

**Expected behavior**
Spectated person doesn't see the one with silent spec enabled.

**Screenshots**
What person spectated sees:
sspechls

What me with silent spec sees and what other spectators see:
sspecme

**Plugin List**
SurfTimer 281 master, SM1.10, MM1.10

**Additional context**
Maybe delete this function/command, not many people use it and it's a bit creepy.

qawery-just-sad:
don't ask why there is so manny commits


LubricantJam:
Removing isn't a solution, it's just evading it. I've got a different version which works, I'll commit that soon.

PR #66, opened 2020-04-26
Replace ColorVariables with ColorLib

Bara:
ColorLib on [Github](https://github.com/c0rp3n/colorlib-sm) and on [AlliedModders](https://forums.alliedmods.net/showthread.php?t=320842)

Reason for ColorLib:

> The reason for this is that Colors and More Colors both make heavy use of many ReplaceString operations each one looping through the buffer being formatted, whereas this could be avoided by writing a single pass formatter as used here and was done for ColorVariables.
And to why replace ColorVariables? it also adds extra overhead with forwards and allowing for more "dynamic" colors, which isnt always needed.

[Source](https://forums.alliedmods.net/showthread.php?t=320842)

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -1,8 +1,6 @@
 name: Compile with SourceMod
 
-on:
-  push:
-    branches: ["master", "githubActions", "dev"]
+on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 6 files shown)


olokos:
`
SurfTimer | Available Colours: darkred lightred red green lime lightgreen darkblue lightblue blue pink purple orange yellow {`
Apparently this gets printed to console after merging it with dev and testing on a live server.

You can try it out on `connect kiepownica.pl`, just built the latest changes on dev branch including colorlib merge.

PR #68, opened 2020-04-26
Update readme

PwnK:
Some links still weren't updated in readme.


olokos:
Thanks, we've missed those!

PR #69, opened 2020-04-26
better sql/stripper files/display center speed after map change/no sound on saveloc/allow making startpos with saveloc

fatt1:
previous default zones are god awful especially for grinding stage records - these are from my db and most of them should be picture perfect. Also added stripper files that should cover most maps.
Also put checker in OnClientPutInServer to automatically display center speed if its enabled, so dont have to manually do it everytime. Making startpos with saveloc alot easier and most people prefer that, also disabled buttonsound when pracmode enabled


fatt1:
> Looks good to me, not sure about all the stripper config, but I would merge this.

Stripper files are nothing too fancy, few lasers on maps that needs them (for some invis stuff), sum fixed teles, few playerspawns and stuff like that. Most of the zones should cover info_tele_destination which was main issue (imo) in previous zones. How that saveloc thingy was made might not be the most elegant solution, but works well and feels pretty unexploitable (ive had it on my live servers for almost a year)

addons/sourcemod/scripting/surftimer/misc.sp
```diff
@@ -4408,6 +4408,48 @@ public void TeleportToSaveloc(int client, int id)
 	DispatchKeyValue(client, "targetname", g_szSaveLocTargetname[id]);
 	SetEntPropVector(client, Prop_Data, "m_vecVelocity", view_as<float>( { 0.0, 0.0, 0.0 } ));
 	TeleportEntity(client, g_fSaveLocCoords[id], g_fSaveLocAngle[id], g_fSaveLocVel[id]);
+	CreateTimer(0.1, noPrac, client);
```
Bara, line 4411: Please use convert the client index to the userid or serial index with `GetClientUserId` or `GetClientSerial` to prevent race conditions.  To get back the client index use `GetClientFromSerial` or `GetClientOfUserId` and check if the client index is still valid.

olokos:
I think we should wait with merging this PR as @KristianP26 is working on zoning the maps by hand 
He will check those currently in the repo, those you've sent in @fatt1 and those he has made.
Lets make sure to get the best zones possible. 
I would merge the rest of the changes you've sent in @fatt11 , we should wait for somebody to approve aswell.

I think it would be good if you merge the other 3 commits with this PR, but make another PR for
ac2b8cc6f481846df55deca63a5718a850a89743

olokos:
I think it will be better to merge this, as some of the current zones are completely broken.
Thank you so much @fatt1 for your contribution! ❤️

PR #70, opened 2020-04-26
Revert "General Timer Improvements"

LubricantJam:
Reverts olokos/Surftimer-olokos#59


olokos:
Why tho? Was good progress.

LubricantJam:
> Why tho? Was good progress.

What? It was a mistake hence it's closure?

PR #73, opened 2020-04-27
Restore Previous Records

LubricantJam:
It seems like this fork of the timer has never had the ability to delete the previous bot file? That's very odd so I've added this as well as when you delete a record, instead of just deleting the .rec and being done with it, it'll restore the previously stored record.

### DIFF
--- a/addons/sourcemod/scripting/surftimer/sql.sp
+++ b/addons/sourcemod/scripting/surftimer/sql.sp
@@ -8696,6 +8696,7 @@ public void db_selectMapCurrentWR(Handle owner, Handle hndl, const char[] error,
 	if (hndl == null)
 	{
 		LogError("[SurfTimer] SQL Error (db_selectMapCurrentWR): %s", error);
+		CloseHandle(pack);
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


addons/sourcemod/scripting/surftimer/sql.sp
```diff
@@ -8675,6 +8691,80 @@ public int ChooseMapMenuHandler(Menu menu, MenuAction action, int param1, int pa
...
+	}
+
+	if (SQL_HasResultSet(hndl) && SQL_FetchRow(hndl))
+	{
+		ResetPack(pack);
+		char szSteamId[32], szStyleType[5];
+		ReadPackString(pack, szSteamId, sizeof(szSteamId));
+		ReadPackString(pack, szStyleType, sizeof(szStyleType));
+		int iLevel = ReadPackCell(pack);
+		CloseHandle(pack);
```
Bara, line 8709: I prefer to close the handle at the end of this callback, outside of `if (SQL_HasResultSet(hndl) && SQL_FetchRow(hndl))` and in `if (hndl == null)`. Otherwise it could leak.

Bara: I already saw this a few times and will update this in a further pull request.

olokos:
So I suppose the PR is not ready for merging yet, or is it? @Bara

Bara:
Feel free to merge it, one potential memory leak more or less shouldn't matter.

olokos:
@Bara so we merge this and you would then send another PR that would take care of the memory leak? :P

PR #74, opened 2020-04-27
Make sure buffer is null terminated at PrintCSGOHUDText

PwnK:
The char array in PrintCSGOHUDText is filled with '\n' characters but the final character must be null.

### DIFF
--- a/addons/sourcemod/scripting/surftimer/misc.sp
+++ b/addons/sourcemod/scripting/surftimer/misc.sp
@@ -4903,8 +4903,10 @@ void PrintCSGOHUDText(int client, const char[] format, any ...)
 	VFormat(buff, sizeof(buff), format, 3);
 	Format(buff, sizeof(buff), "</font>%s ", buff);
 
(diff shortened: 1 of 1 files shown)


sneak-it:
Per @kidfearless who wrote the function:

> the original loop would overwrite the last null terminator and would most likely cause issues down the line

PwnK:
According to [alliedmodders wiki](https://wiki.alliedmods.net/Introduction_to_SourcePawn_1.7#Strings) all strings must end in null character '\0' so it would be good practise to do it with this too although the original function may not have caused any big issues.

Without the null character at end pawn may not stop reading the string at the last '\n' character but continue to read trash from memory until it encouters null character.

sneak-it:
Thanks for the clarification - have you tested this change?

PwnK:
Yes I have. I don't use the dev version of this timer in my server but I use this function in my older version of the timer so it should work fine.

sneak-it:
Thank you for the contribution! 🎉

PR #84, opened 2020-04-30
Split master.yml into release.yml and dev.yml

Bara:
We this step, it's easier for us to build/release a package on the master branch and only on the master branch.


covertxd:
I would have it for release only so we don't have to worry about any Github storage limits.

olokos:
Good point.
I'm not entirely sure, but is this PR going to generate downloadable .smx artifacts on build? 🤔

Bara:
That's why I created a own workflow for the master branch, which we use for releases.

PR #99, opened 2020-05-05
updated the Upgrade files

TheGunter24:
this removes more old surftimer stuff
and checkpoint upgrade has the create index included

### DIFF
--- a/scripts/upgrade_scripts/upgrade-fluffy.sh
+++ b/scripts/upgrade_scripts/upgrade-fluffy.sh
@@ -4,26 +4,26 @@
 #   upgrade script from flufflys timer   #
 ##########################################
 
-# change this to your csgo directory, if not running from "upgrade_scripts" folder
-SERVERDIR="../.."
+# change below to your csgo server root directory
+SERVERDIR="/home/csgoserver/csgo/"
 
-rm -rf "${SERVERDIR}/sound/quake"
-rm -f "${SERVERDIR}/sound/surftimer/*.mp3"
+rm -r "${SERVERDIR}/sound/quake"
+rm "${SERVERDIR}/sound/surftimer/*.mp3"
 mv "${SERVERDIR}/sound/surftimer/pr/valve_logo_music.mp3" "${SERVERDIR}/sound/surftimer/pr.mp3"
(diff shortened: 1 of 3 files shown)


scripts/upgrade_scripts/upgrade-fluffy.sh
```diff
@@ -4,17 +4,26 @@
...
 ##########################################
 
-# change this setting to your install directory
-SERVERDIR="/srv/csgo-server/csgo"
+# change this to your csgo directory, if not running from "upgrade_scripts" folder
+SERVERDIR="../.."
 
 rm -rf "${SERVERDIR}/sound/quake"
-rm "${SERVERDIR}/sound/surftimer/*.mp3"
+rm -f "${SERVERDIR}/sound/surftimer/*.mp3"
```
olokos, line 11: do we really need to force it?, same applies to the other lines in this file in general

olokos:
I would merge this, I am not so sure about using `rm -f` everywhere, as rm -rf is potentially dangerous, even if in correct directories and is considered a bad practice, at least I heard it somewhere. 😄 

Thank you for your help so far!

sneak-it:
Ditto to @olokos sentiment, let's not forcibly delete stuff if we don't need to.

13ace37:
Forcing is def not needed. Also using "../.." is kinda bad.

scripts/upgrade_scripts/upgrade-fluffy.sh
```diff
@@ -4,17 +4,26 @@
 #   upgrade script from flufflys timer   #
 ##########################################
 
-# change this setting to your install directory
-SERVERDIR="/srv/csgo-server/csgo"
+# change this to your csgo directory, if not running from "upgrade_scripts" folder
+SERVERDIR="../.."
```
olokos, line 8: maybe we could do
`# change below to your csgo server root directory
SERVERDIR="/home/csgoserver/csgo/"`
This would've been easier to undertand to people ✌️

olokos:
We're just trying to make everything as perfect as possible, you're doing good work, don't get yourself down! 😄

TheGunter24:
done :D 


> We're just trying to make everything as perfect as possible, you're doing good work, don't get yourself down! 😄

absolutly fine, only way to learn :D

olokos:
Good job, thank you for your help @TheGunter24 !

PR #104, opened 2020-05-07
Change Map Fixes

LubricantJam:
Just a small fix to move away from changelevel as you can sometimes get an issue with the map changing back to the same map. This can be fixed by an external plugin, however this way is a lot cleaner and is integrated into the timer for a natural map change.

### DIFF
--- a/addons/sourcemod/scripting/surftimer/timer.sp
+++ b/addons/sourcemod/scripting/surftimer/timer.sp
@@ -176,9 +176,9 @@ public Action CKTimer2(Handle timer)
 				case 60:CPrintToChatAll("%t", "TimeleftSeconds", g_szChatPrefix, g_szMapName, timeleft);
 				case 30:CPrintToChatAll("%t", "TimeleftSeconds", g_szChatPrefix, g_szMapName, timeleft);
 				case 10:CPrintToChatAll("%t", "TimeleftSeconds", g_szChatPrefix, g_szMapName, timeleft);
-				case 3:CPrintToChatAll("%s {green}3!", g_szChatPrefix);
-				case 2:CPrintToChatAll("%s {orange}2!", g_szChatPrefix);
-				case 1:CPrintToChatAll("%s {red}1!", g_szChatPrefix);	
+				case 3:CPrintToChatAll("%t", "TimeleftThree", g_szChatPrefix);
+				case 2:CPrintToChatAll("%t", "TimeleftTwo", g_szChatPrefix);
+				case 1:CPrintToChatAll("%t", "TimeleftOne", g_szChatPrefix);	
 				case 0:
 				{
 					if (!g_bRoundEnd)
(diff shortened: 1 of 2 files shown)


olokos:
But at this point we should provide users with some repository for a mapchooser, or will the ingame vote system work for this aswell, if no mapchooser plugin is in place?

dPexxIAM:
Mapchooser made by fluffys works great with SurfTimer. Ace made a fork of it which removed all the hardcoded stuff among other small fixes.

If any map chooser should be shipped with the timer, or at least recommended it should be this one: https://github.com/13ace37/surftimer-mapchooser

LubricantJam:
This will work for the in-game map chooser too? Timer simply just ends the round, nextmap is set by whatever map chooser you use, this simply just ends the round.

olokos:
It does work well, I am running it on my server and can confirm.

before doing this though, we need a script, which will convert csgo/maps/%surf% to SQL Query, so all maps from the server are added to the mapchooser.

LubricantJam:
> It does work well, I am running it on my server and can confirm.
> 
> before doing this though, we need a script, which will convert csgo/maps/%surf% to SQL Query, so all maps from the server are added to the mapchooser, as currently there will be maps that are not on the server, which will cause problems and does sometimes.

Should be done in that repo or a different fork to work along side this. Safe to merge, I've tested on my own servers.

olokos:
I know, just wanted to make sure it's known.
@13ace37 Could you please update the readme of your mapchooser with the python script. please? ☺️

addons/sourcemod/scripting/surftimer/timer.sp
```diff
@@ -176,18 +176,15 @@ public Action CKTimer2(Handle timer)
 				case 60:CPrintToChatAll("%t", "TimeleftSeconds", g_szChatPrefix, g_szMapName, timeleft);
 				case 30:CPrintToChatAll("%t", "TimeleftSeconds", g_szChatPrefix, g_szMapName, timeleft);
 				case 10:CPrintToChatAll("%t", "TimeleftSeconds", g_szChatPrefix, g_szMapName, timeleft);
-				case 3:CPrintToChatAll("%s ~~~ MAP ENDING ~~~", g_szChatPrefix);
-				case 2:CPrintToChatAll("%s ~~~ MAP ENDING ~~~", g_szChatPrefix);
-				case 1:CPrintToChatAll("%s ~~~ MAP ENDING ~~~", g_szChatPrefix);
-				case 0:CreateTimer(14.0, ForceNextMap, INVALID_HANDLE, TIMER_FLAG_NO_MAPCHANGE);	
-				case -1:
+				case 3:CPrintToChatAll("%s {green}3!", g_szChatPrefix);
```
olokos, line 179: Maybe we could change this to Map ending in 3! and so on? Might be more obvious what is going on in the future! 😄

Bara:
@olokos or fork it and make it yourself.

addons/sourcemod/scripting/surftimer/timer.sp
LubricantJam: I think it's personally cleaner this way, as you've got the multiple second warnings above. No one has ever second guessed what it is, if you want to add translations for it, go ahead. I just wanted the fix in to keep the dev branch stable.

olokos:
@Bara I think he's doing a good job with the repo and if he doesn't mind adding it, I shouldn't fork it just yet.

addons/sourcemod/scripting/surftimer/timer.sp
olokos: I think just keeping it as `Map restarting in 3` would be enough, no need to put in in the translations just yet, we can put a comment there about possibly adding it in the future to translations.

If you don't mind, I think it would be useful for those that don't know the code that good.

LubricantJam: > I think just keeping it as `Map restarting in 3` would be enough, no need to put in in the translations just yet, we can put a comment there about possibly adding it in the future to translations.
> 
> If you don't mind, I think it would be useful for those that don't know the code that good.

I disagree but yeah, too much text.

PR #105, opened 2020-05-08
Create german translation

TheGunter24:
made a german translation
was more work then first believed :D 
not perfect yet, feel free to give feedback


Bara:
Im Grunde brauchst du die Zeilen mit `#format` nicht mit reinpacken, dies wird von der englischen Übersetzung übernommen.  Als Beispiel wären die Übersetzungen von SourceMod.

You don't need to add the lines `#format`, this part are just required in the english translations. As example are the default translations by sourcemod.

TheGunter24:
> You don't need to add the lines `#format`, this part are just required in the english translations. As example are the default translations by sourcemod.

ehh, ok, but it doesnt hurt right ?
would be some work :D

Bara:
As long as it works, but in future changes it must be updated too.

PwnK:
In case you dont want to delete every `#format` line by hand you can use this regular expression in notepad++ and replace it with nothing `\t\t\"#format\".*\r\n`

TheGunter24:
> In case you dont want to delete every `#format` line by hand you can use this regular expression in notepad++ and replace it with nothing `\t\t\"#format\".*\r\n`

you are absolutly right ^^
i can use notepad tools, thx

TheGunter24:
ive changed the "fr" trans too

TheGunter24:
is ready to merge if someone would pls do it ^^
(just updated to fit match https://github.com/olokos/Surftimer-olokos/commit/87606996a0f8052c2d7c8cacf769a833f2988396)

Issue #76, opened 2020-04-28
[BUG]BOT Reply bug


Xedoo:
**Describe the bug**
Example : first guy in top have map record 2 min and 23 seconds and 12 ms ok? The another have a different time. When you want to see the bot reply 1 time bot will be on first guy and second time bot will be an another guy different time. In TAB appear a random guy with random time not the First guy with the best time. After sm_fixbots the problem is the same. Map record and bonus same problem. 

**Additional context**
Another thing, when i use spec on bot or sometimes on players, when i want to play again, !r or !back or !start, any command to 'revive' sometimes the screen goes on black and i need to reconnect to the server to play again. Not happen only for me.

EDIT : BOT Replay for stages is like finish the same stage 3 times and after the next stage again 3 times...

TheGunter24:
could you try to rewrite it ? cant barely understand what you want
some screenshots would be awesome

regarding your edit:
stage replays are ment to run 3 times â stage

Kyli3Boi:
I think what he is trying to say is the same issue that i have as well as BlazerV60 (on discord)

Our replay bots are not playing the server record replay but rather some other random player.
IE bot is replaying a run of a person ranked 7th,10th,20th rather than the person that is ranked 1st.

Fixbot does nothing as it just loads the same bot.

I even have a bot replaying a run of a person that starts, gets half way, fails, starts again and then completes and the bot is replaying all of this

LubricantJam:
@fluffyst I believe has a fix for this, he would have to merge it.

Kyli3Boi:
@LubricantJam I spoke with fluffyst and he says that the solution you thinking of is not for the same problem.

Any other suggestions? Anything i can try do?

LubricantJam:
> @LubricantJam I spoke with fluffyst and he says that the solution you thinking of is not for the same problem.
> 
> Any other suggestions? Anything i can try do? 
> 

No idea, will be a case of figuring out the problem I guess.

Xedoo:
> 
> 
> I think what he is trying to say is the same issue that i have as well as BlazerV60 (on discord)
> 
> Our replay bots are not playing the server record replay but rather some other random player.
> IE bot is replaying a run of a person ranked 7th,10th,20th rather than the person that is ranked 1st.
> 
> Fixbot does nothing as it just loads the same bot.
> 
> I even have a bot replaying a run of a person that starts, gets half way, fails, starts again and then completes and the bot is replaying all of this

[...]

PR #123, opened 2020-05-13
Minor Fixes

LubricantJam:
Here's a couple few minor fixes.

- Fixed fun styles being set back to normal movement after Command_Teleport action is called.
- Potential fix for the replay bug issue #76 but I am not too sure.
- Fix for the Spectator HUD sometimes glitching out and going off-center slightly and then moving back as well as changed back to the nicer colour format.


sneak-it:
I think this is a bit of a hacky fix for the issue - at the core how player gravity/m_fflaggedmovement is applied needs to be reworked. This seems like an acceptable temp fix, so I'll +1 the PR

PR #136, opened 2020-05-16
Update README.md

erikjunker:
Made it more obvious that colorlib_map.inc is not needed from that repository


erikjunker:
My bad - that commit has been reverted now.

PR #148, opened 2020-05-22
[DataPack] Fix few memory leaks and a few more race conditions

Bara:
This is definitely a few breaking changes, if something went wrong. So please test, before review.

### DIFF
--- a/addons/sourcemod/scripting/surftimer/hooks.sp
+++ b/addons/sourcemod/scripting/surftimer/hooks.sp
@@ -1354,7 +1354,7 @@ public Action Event_PlayerJump(Handle event, char[] name, bool dontBroadcast)
 				CPrintToChat(client, "%t", "Hooks10", g_szChatPrefix);
 				Handle pack;
 				CreateDataTimer(0.05, DelayedVelocityCap, pack);
-				WritePackCell(pack, client);
+				WritePackCell(pack, GetClientUserId(client));
 				WritePackFloat(pack, 0.0);
 				g_bJumpZoneTimer[client] = true;
 			}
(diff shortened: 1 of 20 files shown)


addons/sourcemod/scripting/surftimer/hooks.sp
```diff
@@ -1457,23 +1457,27 @@ public Action ResetOneJump(Handle timer, any userid)
 public Action DelayedVelocityCap(Handle timer, Handle pack)
 {
 	ResetPack(pack);
-	int client = ReadPackCell(pack);
+	int client = GetClientOfUserId(ReadPackCell(pack));
 	float speedCap = ReadPackFloat(pack);
+	delete pack;
```
PwnK, line 1462: DelayedVelocityCap is a timer callback which is called by [`CreateDataTimer`](https://sm.alliedmods.net/new-api/timers/CreateDataTimer) so we don't need to close the datapack which comes as it's parameter.

Before making changes that close datapacks in timer callback parameters I recommend to always check if it's called by [`CreateTimer`](https://sm.alliedmods.net/new-api/timers/CreateTimer) or [`CreateDataTimer`](https://sm.alliedmods.net/new-api/timers/CreateDataTimer).

Bara: Yeah I remember... 🤦‍♂️
https://github.com/surftimer/Surftimer-olokos/pull/103#pullrequestreview-407890736

I changed that to CreateTimer now.

olokos:
I am not really sure why change all CreateDataTimer to CreateTimer though.
https://github.com/alliedmodders/sourcemod/issues/682

Given above issue, CreateDataTimer would be the preffered method to call, there's less complexity later when those methods are changed/removed as you don't have to manually close all handles, it's done automatically with CreateDataTimer.

Correct me if I'm wrong here, but I see no upsides of switching to CreateTimer

olokos:
Merged, but with this change 
e868c8ec9530e082849a891ae75e5b0356d91b87
delete handles were reintroduced, do we really need them if we're only using CreateDataTimer?
@Bara 
Let me know your take on this, should I create an issue so we can tackle this later on?

PR #149, opened 2020-05-22
Fix some issues

Bara:


olokos:
I would appreciate if somebody else could also take a look on this and approve this, before merging though.

PR #159, opened 2020-05-27
Changed the discord announcements into nicer embed

Sarrus1:
Changed the discord announcements into nicer embeds messages. I will post a screenshot on the Discord


Sarrus1:
Yep, the picture I've uploaded on the discord is from when compiling on v285 with those changes. It works perfectly.
I've talked with others and they agree with me that we should add some cvars to allow some form of customization of the way the embed is displayed. I'll try to work on that in the coming days :)

PR #161, opened 2020-05-29
Revamped the discord messages [Tested and working]

Sarrus1:
Ok here is the new version of the embeds, for the bonus and the regular records:

### What's new:

**There new CVars:**
- Choose between the KSF style or the embeds.
- Choose who is mentioned on discord, set "" to disable it.
- The title of the embed can be set, bonus and global records separatly.
- You can choose the color of the embed.
- You can disable the discord testing command.

**There is a new command:**
- `sm_ck_discord_test` which sends test messages to discord. It can be disabled with a convar to prevent spam by other players.

**Added a translation:**
- Added a phrase to notify the player that the discord test has been sent.


aKjeller:
Would be nice if the messeage also include the time improvment from the last SR!

Sarrus1:
True, I'm not sure how to do that however :/ I guess adding another parameter to the function would work

> Would be nice if the message also include the time improvment from the last SR!

Sarrus1:
Sorry for the spam, I'm still unfamiliar with GitHub :/


> Would be nice if the message also include the time improvment from the last SR!

I've added that in the latest commit :)

PR #186, opened 2020-06-18
Fix discord msg

qawery-just-sad:
Not tested. (I don't have time for that, plz someone test that.)
Time should be compared to previous SR.


qawery-just-sad:
Almost forgot, how is my favourite test subject doing @FR33K1LL ?

qawery-just-sad:
nevermind it doesn't work

PR #187, opened 2020-06-19
Update surf_mesa.cfg

Kyli3Boi:
Added filter to remove map sounds, which was spamming the console with multiple errors

### DIFF
--- a/addons/stripper/maps/surf_mesa.cfg
+++ b/addons/stripper/maps/surf_mesa.cfg
@@ -2,9 +2,6 @@ filter:
 {
   "classname" "env_sun"
 }
-{
-  "classname" "env_soundscape"
-}
 
 add:
 {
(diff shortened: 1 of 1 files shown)


addons/stripper/maps/surf_mesa.cfg
```diff
@@ -137,3 +137,8 @@ add:
   "classname" "env_laser"
   "hammerid" "3875"
 }
+
+filter:
```
sneak-it, line 141: Just nitpicking, you can add this filter to the top of the file after the env_sun removal for cleanliness.

Like so:
```
filter:
{
  "classname" "env_sun"
}
{
  "classname" "env_soundscape"
}
```

Kyli3Boi:
Sorry for that Disaster, im still not confident in how github works.

I have taken your suggestion and amended

Kyli3Boi:
If there is a way i can clean up all those commits please let me know

sneak-it:
Looks great, thanks for your contribution!

PR #191, opened 2020-06-26
Fix beyond styles

qawery-just-sad:
The following stripper config allows players to use any styles suported by plugin.


sneak-it:
Uhh, why are you deleting the vscript that nulls speed and stops the ability to telehop?

qawery-just-sad:
erm.. can't stage speed limiter handle that? @sneak-it

sneak-it:
> erm.. can't stage speed limiter handle that? @sneak-it 

That's not how that works at all. Why was this removed?

PR #198, opened 2020-07-02
Fix discord msg

qawery-just-sad:
The discord messages will now be compared to SR
tested but as @sneak-it said 2 approvals are needed and i want someone else to test it as well

### DIFF
--- a/addons/sourcemod/scripting/surftimer/buttonpress.sp
+++ b/addons/sourcemod/scripting/surftimer/buttonpress.sp
@@ -370,6 +370,7 @@ public void CL_OnEndTimerPress(int client)
 					CreateTimer(0.0, ReplayTimer, GetClientUserId(client), TIMER_FLAG_NO_MAPCHANGE);
 				}
 
+				g_fOldRecordMapTime = g_fRecordMapTime;
 				g_bMapSRVRecord[client] = true;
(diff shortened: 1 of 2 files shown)


Kyli3Boi:
Im currently testing 006d817. Will give feedback once i have collected enough data

Kyli3Boi:
Still seems to be the same but as you guys have pointed out before could be due to the version im using...
image

qawery-just-sad:
strange 
image

Kyli3Boi:
I think your testing method is incorrect.

It works perfectly fine when it's the same player that beat the record as you have already finished the map. The issue is when a different player finishes the map/bonus for the 1st time and beats the record, that's when it bugs out

Kyli3Boi:
Just Sad and myself have tested this out and it working as intended on his server. Unsure as to why mine is not working as intended but could be a version issue or something. 

But this seems good to merge as its working as intended (at least on the version Just Sad is using)

stunslave:
Can you make it so it shows the tier of the map next to the map name. :P

Map
surf_frost (Tier2)

qawery-just-sad:
that wasn't a question was it?

olokos:
Thank you for your contribution! 👍

PR #204, opened 2020-07-08
Should fix #133

### LINKED ISSUE #133: [BUG] Hook Zoning Exception
**Describe the bug**
When zoning with a Hook Zone, you are only able to use this option once and any attempts after will give you a handle error and the menu will close upon pressing "4. Hook Zone". Retrying allowed me to do it again but eventually stopped working even after retrying. 

**Expected behavior**
Ability to use Hook Zone multiple times

**Screenshots**
https://streamable.com/8jbgyz

**Plugin List**
List of all your current sourcemod plugins, if applicable to the issue.

**Timer Version**
285.9

**Additional context**
```
[SM] Exception reported: Menu handle ab1802a0 is invalid (error 1)
[SM] Blaming: SurfTimer.smx
[SM] Call stack trace:
[SM]   [0] DisplayMenu
[SM]   [1] Line 4292, surftimer/commands.sp::HookZonesMenu
[SM]   [2] Line 1601, surftimer/surfzones.sp::Handle_SelectNormalZoneType
```

qawery-just-sad:
NOT TESTED
@FR33K1LL could you test it?

### DIFF
--- a/addons/sourcemod/scripting/surftimer/commands.sp
+++ b/addons/sourcemod/scripting/surftimer/commands.sp
@@ -4497,6 +4497,8 @@ public int HookZonesMenuHandler(Menu menu, MenuAction action, int param1, int pa
 		}
 		case MenuAction_End:
 		{
+			if (IsValidClient(param1))
+				g_iSelectedTrigger[param1] = -1;
 			delete menu;
 		}
 	}
--- a/addons/sourcemod/scripting/surftimer/commands.sp
+++ b/addons/sourcemod/scripting/surftimer/commands.sp
@@ -4495,12 +4495,6 @@ public int HookZonesMenuHandler(Menu menu, MenuAction action, int param1, int pa
 			if (IsValidClient(param1))
 				g_iSelectedTrigger[param1] = -1;
 		}
-		case MenuAction_End:
-		{
-			if (IsValidClient(param1))
-				g_iSelectedTrigger[param1] = -1;
-			delete menu;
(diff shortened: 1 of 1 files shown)


PwnK:
Param1 is not client index in MenuAction_End so this change is just wrong. See https://wiki.alliedmods.net/Menu_API_(SourceMod)

What you should do is just remove the MenuAction_End altogether because it is closing the menu handle which is not created again every time the command is run. g_mTriggerMultipleMenu is created once every time the map starts and it doesn't need to be closed every time the command is run.

qawery-just-sad:
then please explain me that as well
https://github.com/surftimer/Surftimer-olokos/blob/3f1cd2ab7058806339f5bb1cd175fc55c3a01c1c/addons/sourcemod/scripting/surftimer/commands.sp#L4554

PwnK:
https://github.com/surftimer/Surftimer-olokos/blob/3f1cd2ab7058806339f5bb1cd175fc55c3a01c1c/addons/sourcemod/scripting/surftimer/commands.sp#L4553-L4559

Yeah this is also wrong and it should be removed from MenuAction_End. The same thing is done in MenuAction_Cancel where param1 is client index.

qawery-just-sad:
Thank you @PwnK

qawery-just-sad:
Pretty much yes

FR33K1LL:
Sweet no need to test then, thats the fix i did in may, and been using ever since.

PR #205, opened 2020-07-08
removed enoing music.

KristianP26:


sneak-it:
Seems pretty trivial, I don't think we should be changing anything that's not strictly a fix. If server owners want to remove music from maps, they are free to do so on their own accord.

PR #206, opened 2020-07-08
Multiple stripper updates

Kyli3Boi:
Apologies this is the same as Multiple stripper updates #202. Just having some GitHub issues xD


olokos:
@sneak-it  can you check this out? You know stripper better.

Issue #53, opened 2020-04-23
[BUG] Using any !style ignores the zone max prespeed


breakthegreat:
**Describe the bug**
Using any style will ignore the zone prespeed limit, attached a screenshot of low gravity still where I can go over 350 u/s limit in the start zone
**Expected behavior**
Style should not ignore start zone prespeed limits. 
**Screenshots**
![image](https://user-images.githubusercontent.com/64219376/80136330-8bb54c00-856f-11ea-892a-17788c084ed0.png)

**Additional context**
Was an issue back in the older version of surftimer (274).

13ace37:
This is not 100% true. Also, you could simply enable the one jump limit for the zones.

breakthegreat:
I know but one jump limit is somewhat weird for people who plays on the server, they don't really like it, but definetely i'd put this one jump limit if it can't be fixed

breakthegreat:
To be more specific, this bug only appears with fun styles (ranked seems to work fine)

sneak-it:
If I'm understanding correctly, you want a hard start zone speed cap, if so use `ck_limit_speed_type 0`.

breakthegreat:
ck_limit_speed_type did the trick, i'm assuming I should put this on from now on, but surftimer way to limit speed is so much better. Idk if this style issue will be fixed anytime soon, but doesn't really matter as much as it is mostly affecting fun styles.

olokos:
Is `ck_limit_speed_type 0` going to completely disable speed limiter?
I'd love some more insight into it, please.

freakexeuLow:
I have a similar problem, maybe it'll help...here a small video where a user leaves the zone with a prespeed of 500 speed :3
the value in cfg is set to: ck_limit_speed_type "1"

[https://youtu.be/2EjATUBA628](https://youtu.be/2EjATUBA628)

sneak-it:
Not sure how this is an issue... the way `ck_limit_speed_type 1` works is by only limiting speed if a bhop is performed. Legitimate prestrafing on the ground and in the air (without bhop) is uncapped.

freakexeuLow:
> 
> 
> Not sure how this is an issue... the way `ck_limit_speed_type 1` works is by only limiting speed if a bhop is performed. Legitimate prestrafing on the ground and in the air (without bhop) is uncapped.

Hey sneaK :3 
good to know, many of the users write with me that it is abusing and i was thinking the prespeed is normal capped 350 everytime ^^

[...]

PR #208, opened 2020-07-10
"Fixes" #129 and #53

### LINKED ISSUE #129: [BUG] Low grav / Slow mo times glitch
**Describe the bug**
If you set your style to low grav or slow mo, then switch teams, it will keep you in the chosen style but anything the style changed (speed, gravity etc) will be reset to normal. Your time will still be set for the chosen style (lg or sm) but with the style settings for normal style. 

**Expected behavior**
When teams are switched, the chosen styles settings should remain applied and not reset to normal.

**Screenshots/Video**
[Link to video](https://streamable.com/y5m2bv)

**Plugin List**
List of all your current sourcemod plugins, if applicable to the issue.

**Timer Version**
dev 285.7

**Additional context**
Bug is also present in 285. I assume this affects all styles, but these 2 styles are the only ones that slow your speed.

qawery-just-sad:
Yes it Fixes #129  BUT with #53 there is a copule of little problems:

 - New issue, if you use `ck_one_jump_limit 1` while in low-grav or slowmo you are not able to to move and it spams chat with "You can only jump once in start zone" even though you are not jumping.

- New issue if in lg/sm your speed gets capped by zone prespeed setting and does not allow prestrafing like in normal style

### DIFF
--- a/addons/sourcemod/scripting/surftimer/hooks.sp
+++ b/addons/sourcemod/scripting/surftimer/hooks.sp
@@ -939,8 +939,170 @@ public Action OnPlayerRunCmd(int client, int &buttons, int &impulse, float vel[3
 		else if (!g_bInStartZone[client] && !g_bInStageZone[client] && val < -0.75)
 		g_KeyCount[client] = 0;
 	}
+	else if (g_iCurrentStyle[client] == 4) // Low-gravity
+	{
+		if (g_bInStartZone[client] || g_bInStageZone[client])
+		{
+			ResetGravity(client);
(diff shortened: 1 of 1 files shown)


qawery-just-sad:
if someone is interested i would need help

qawery-just-sad:
So there is the fix for #129. The fix for #53 will require rewrite of prehopping/strafing in start zones so i'll leave it for some more experienced people.

Kyli3Boi (approved):
Tested and working

PR #216, opened 2020-08-01
Fix fixbot command

sneak-it:
Fix duplicate and breaking fixbot permissions check, allow to be used via server console

### DIFF
--- a/addons/sourcemod/scripting/surftimer/commands.sp
+++ b/addons/sourcemod/scripting/surftimer/commands.sp
@@ -4217,10 +4217,7 @@ public Action Command_Repeat(int client, int args)
 
 public Action Admin_FixBot(int client, int args)
 {
-	if (!g_bZoner[client] && !CheckCommandAccess(client, "", ADMFLAG_ROOT))
-		return Plugin_Handled;
-
-	CPrintToChat(client, "%t", "Commands52", g_szChatPrefix);
+	CReplyToCommand(client, "%t", "Commands52", g_szChatPrefix);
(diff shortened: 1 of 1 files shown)


Kyli3Boi:
Thanks sneak

sneak-it:
> As far as I'm aware there are checks being done on client before this command can even be called, so this nested if statement is possibly reduntant.

The permission assigned to it is being checked in the registered command - the nested checks were not only redundant, but also horribly hard-coded and didn't allow using SourceMod's overrides system properly (as discussed in Discord).

qawery-just-sad:
the same is true for all vip commands (just saying)

sneak-it:
[**@qawery-just-sad**](https://github.com/qawery-just-sad) commented on [Aug 2, 2020, 6:01 AM EDT](https://github.com/surftimer/Surftimer-olokos/pull/216#issuecomment-667653777 "2020-08-02T10:01:36Z - Replied by Github Reply Comments"):
> the same is true for all vip commands (just saying)

Oh gross, that "IsPlayerVip" stock needs to be nuked from orbit. Looks like that's used on a lot of VIP commands, not so much the case with fixbot, which is outside the scope of this PR.

olokos:
@sneak-it 
Feel free to self-approve this if you think this PR can be closed already.
Possibly a good idea would be to create a new issue when this PR is closed, just to remember about the vipCommands/removal of vipCommands.

dPexxIAM (commented):
lgtm

PR #218, opened 2020-08-11
Add option to log queries

Bara:


sneak-it:
Did you mean to make a convar switch for this, or just a hard coded bool change?

Bara:
I made it configurable per convar.

LubricantJam:
Very useful, can be used to debug queries as said. I'm not too sure if it'll affect the IO too much but it's good enough to warrant the debugging it provides.

PR #219, opened 2020-08-12
Fix incorrect percentage value

Bara:


LubricantJam:
Thanks for this, I'll merge from my phone.

PR #220, opened 2020-08-12
Replace old DBI style with DBI methodmap

Bara:
I've also added a stock to check is a (old hndl) db is valid and error doesn't longer as 1.


Kyli3Boi:
ty for the work bara!

PR #221, opened 2020-08-13
Update upgrade to support ports from older ckSurf versions

Bara:
This version of upgrader supports older ckSurf version which need `upgrade-niko.sql` as example.

I removed the checkpoints upgrade part due issues and memory leaks, we provide the upgrade sql file [here](https://github.com/surftimer/Surftimer-olokos/blob/updateUpgrader/scripts/mysql-files/upgrading/upgrade-checkpoints.sql).

So it isn't automatic anymore, but it is now more safer to use since we run into some performance issues with sourcemod and Transactions with a few thousends of queries.


Bara:
unknown
Seems good for me, error is already fixed [here](https://github.com/surftimer/Surftimer-olokos/pull/221/commits/3a6253685f83f3686f986f73533000ca489921b5#diff-8242ba7df8bce76bc3c4ed9f39df8bc8R80).

PR #230, opened 2020-08-18
Several sql changes

Bara:
Rename "data" to "better" names which makes more sense and make code more readable
Adding missing `DBPrio_low`


Kyli3Boi:
Tested and can confirm this does fix the memory leak issue with #222

Bara:
https://github.com/surftimer/Surftimer-olokos/commit/233529775283f3ba1f9ec42292aa71d2a1d40087#commitcomment-41691574

Issue #155, opened 2020-05-25
[BUG] sm_hide not working on bots


stunslave:
**Describe the bug**
Cant hide bots - All bots are visible (sm_hide only hides players).

**Screenshots**
https://i.imgur.com/SPpJbtg.png

**Timer Version**
285.95 (latest dev)

sneak-it:
Can you or someone else test and see if this happens on the latest commit of the dev branch? @stunslave

Bara:
hidebot branch or not? 🤔

FR33K1LL:
I can test in a couple days if no one else does by then.

sneak-it:
[**@Bara**](https://github.com/Bara) commented on [Aug 20, 2020, 8:31 PM EDT](https://github.com/surftimer/Surftimer-olokos/issues/155#issuecomment-677972857 "2020-08-21T00:31:23Z - Replied by Github Reply Comments"):
> hidebot branch or not? 🤔

Nope, dev :)

Bara:
https://github.com/surftimer/Surftimer-olokos/commits/dev
🤔

PR #239, opened 2020-08-29
Fix for #155

Bara:


LubricantJam (approved):
Don't see any reason why this wouldn't work.

PR #240, opened 2020-08-29
Make discord as optional feature

Bara:


Kyli3Boi:
image

Kyli3Boi:
Sorry for posting this twice but i thought it was me making the error but i got everything running with the latest dev and then used the optional discord branch and replaced the scripting and translations folders from dev with optional discord. Compiled SurfTimer and SurfTimer-Discord and then i get this error again
image

sneak-it:
This PR fixes it: https://github.com/Deathknife/sourcemod-discord/pull/24

Kyli3Boi:
Thanks sneak-it that does solve that issue. 

Just a question SurfTimer-Discord still requires Discord-API in order to function. Is that how its supposed to be?

sneak-it:
> Thanks sneak-it that does solve that issue.
> 
> Just a question SurfTimer-Discord still requires Discord-API in order to function. Is that how its supposed to be?

Ideally Discord-API is only a compile-time requirement, not a requirement to run the timer. After updating your Discord API include files with the ones from the above linked PR and recompiling, do you still encounter issues running the timer without Discord API?

Kyli3Boi:
Yes that is correct.
image

sneak-it:
> Yes that is correct. [EDIT] what i mean is that SurfTimer loads fine but the Surftimer-Discord does not load unless Discord API.smx is loaded also.
> image

Seems reasonable and fairly intentional, you cannot run the Discord module without Discord-API.

Kyli3Boi:
I see thanks for clearing that up just wanted to make sure.

Bara:
That was the reason for "SurfTimer_discord.smx", without the PR by @sneak-it you can't use Discord-API as optional plugin. By creating "SurfTimer_discord.smx" you can use SurfTimer without Discord-API, so it's more an workaround.

Kyli3Boi:
Understood however i still was not able to get SurfTimer_discord.smx to work without using the above PR from sneak-it

Bara:
Try it with this commit

Kyli3Boi:
Yep that seems to work Bara, SurfTimer-Discord is now loading without having Discord API loaded also

Kyli3Boi:
Just to clarify if someone else reads over this. 

SurfTimer-Discord will load without errors, without having Discord API also loaded however it will not work without Discord API being loaded as returns Native not bound errors

Bara:
I'll rework this today or in the next days. Deathknife merged @sneak-it PR and we can do this without SurfTimer_discord now.

PR #244, opened 2020-09-04
this fixes big mistake in #241

qawery-just-sad:


Kyli3Boi:
The above does fix the issue with "missing PB sound" being played 0.1 seconds after start, as well as time going red. BUT only fixes it for Normal style. All other styles still have the issue.

qawery-just-sad:
closed due to @Bara #250

PR #250, opened 2020-09-21
Merge normal style variables

Bara:


Bara:
[SurfTimer.zip](https://github.com/surftimer/Surftimer-Official/files/5257120/SurfTimer.zip)
Precompiled smx

Kyli3Boi:
Tested and can confirm sound issue is fixed in normal style

PR #252, opened 2020-09-23
Fix for issue #245

### LINKED ISSUE #245: [BUG] !mrank not functioning correctly
**Describe the bug**
When using !mrank it does not call correct number of ranks as well as the time.
eg. Kyli3_Boi is ranked 2/0 with a time of  on surf_beginner

**Expected behavior**
Call correct number of ranks and time.

**Screenshots**
N/A

**Plugin List**
N/A

**Additional context**
SurfTimer (1.1.0-dev)

Bara:


Kyli3Boi (approved):
Tested with Bara and can confirm working

PR #255, opened 2020-09-26
Some replay changes

Bara:
Here's the precompiled version:
[SurfTimer.zip](https://github.com/surftimer/Surftimer-Official/files/5287282/SurfTimer.zip)


Kyli3Boi:
Tested and was not recording replays, also after not recording replay it stopped the timer from starting. Let Bara know these issues

Kyli3Boi:
Tested with changes and replay is being saved now. Timer is starting after saving replay.
When beating record new replay is being saved, old replay is backed up as per normal

ck_stop_recording_missed_pb
"0"
Set map time with alternate player (so run is not saved because of beating SR)
Remove replay
Set time with missed pb - replay saved 

"1"
Set map time
Remove replay 
Set time with missed pb - no replay saved

Verdict: ck_stop_recording_missed_pb **functioning correctly** 

ck_stop_recording_time (changed to 300 for testing purposes)
Removed replay
Start run
Waited 350 seconds 
Completed run - replay saved (not sure if this is because SR was set and this conflicted in some way)
Let Bara know about this

Set map time with alternate player (so run is not saved because of beating SR)
Removed replay
Start run
Waited 350 seconds
Completed run -  replay saved (this shouldn't be saved)

Verdict: ck_stop_recording_time **not functioning correctly**

ck_enable_player_recording
I dont fully understand the need for this as it has a default of 0 and maximum of 1
if 0 there will be no players on the server anyways so wont record anything anyways
if 1 there will be a player on server so will record

ck_player_record_limit 
I cant test this as i would need more than 1 player on my test server and seeing as my test server is local this is not possible.

addons/sourcemod/scripting/surftimer/replay.sp
```diff
@@ -74,6 +74,10 @@ public void StartRecording(int client)
 {
 	if (!IsValidClient(client) || IsFakeClient(client))
 		return;
+	
+	StopRecording(client);
```
olokos, line 78: are you sure this is a good idea?

I think only the return is within the if condition above and as it is now it will stop recording as soon as it starts recording everytime.

Issue #79, opened 2020-04-28
[Suggestion] Add ConVar to disable droping knife after typing knife.


KristianP26:
[Suggestion] Add ConVar to disable droping knife after typing knife.

LubricantJam:
Would you like to give it a go yourself in a PR? Could be nice to slowly teach yourself how to become a contributor. Of course, we will review it for you :)

TheGunter24:
If you want to disable knife drop completely -> https://github.com/olokos/Surftimer-olokos/blob/dev/cfg/sourcemod/surftimer/main.cfg#L27

KristianP26:
@TheGunter24 I mean when you type !knife you get a knife and if you already have one they will stack.

KristianP26:
And its buggy when you use plugin for knifes and skins.

sneak-it:
> @TheGunter24 I mean when you type !knife you get a knife and if you already have one they will stack.

This is an easy fix to add a GetPlayerWeaponSlot check on the knife slot to ensure there is no existing weapon there: https://github.com/olokos/Surftimer-olokos/blob/dev/addons/sourcemod/scripting/surftimer/commands.sp#L4047

> And its buggy when you use plugin for knifes and skins.

Sounds like more of an issue with the plugin you use for knives/skins, we won't support that here.

KristianP26:
nah thats inside of surftimer, because that plugins dontdrop you knife.

Kyli3Boi:
This can be closed...

PR #262, opened 2020-10-03
Improvement for #79

Kyli3Boi:
This stops you from "stacking" knives.

If you already have a knife and use !knife you will get a print to chat saying you have a knife.

If you do not have a knife you will be given one.

### DIFF
--- a/addons/sourcemod/scripting/surftimer/commands.sp
+++ b/addons/sourcemod/scripting/surftimer/commands.sp
@@ -4153,9 +4153,9 @@ public Action Admin_FixBot(int client, int args)
 
 public Action Command_GiveKnife(int client, int args)
 {
-	if (IsClientObserver(client))
+	if (IsClientObserver(client) || !IsPlayerAlive(client))
 	{
-		CPrintToChat(client, "%t", "SpectatingKnife", g_szChatPrefix);
+		CPrintToChat(client, "%t", "CannotRequestKnife", g_szChatPrefix);
 	}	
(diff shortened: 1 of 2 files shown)


addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -4153,11 +4153,15 @@ public Action Admin_FixBot(int client, int args)
 
 public Action Command_GiveKnife(int client, int args)
 {
-	if (IsPlayerAlive(client))
+	if (IsPlayerAlive(client) && (GetPlayerWeaponSlot(client, CS_SLOT_KNIFE) != -1))
 	{
-		GivePlayerItem(client, "weapon_knife");
+		CPrintToChat(client, "%t", "AlreadyHasKnife", g_szChatPrefix);
```
qawery-just-sad, line 4158: eeeee i think here it should give the weapon

addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -4153,11 +4153,15 @@ public Action Admin_FixBot(int client, int args)
...
 	}
-
+	else
+	{
+		GivePlayerItem(client, "weapon_knife");
```
qawery-just-sad, line 4162: and here should be the message

addons/sourcemod/scripting/surftimer/commands.sp
Kyli3Boi: It doesnt work that way because != -1

Entity index on success, -1 if no weapon existed.

qawery-just-sad (approved):

Bara (approved):

Issue #86, opened 2020-05-01
[BUG] One Jump limit Bug


KristianP26:
https://www.youtube.com/watch?v=add5fJYLRQ8&feature=youtu.be

When you jumping and jumping in start zone one jump limit disappear.

Bara:
Are you able to reproduce this with the latest dev Version? If you get any errors, please update or post the errors again.

Don't forget **after updating** surftimer to run the command `sm_surftimer_upgrade` and after this to [.sql](https://github.com/surftimer/Surftimer-Official/blob/dev/scripts/mysql-files/optimizations/ck_change_column_types.sql)-File.

Kyli3Boi:
Am able to reproduce

"SurfTimer" (1.1.0-dev)

PR #263, opened 2020-10-03
One_jump_limit-Band-aid

Kyli3Boi:
This is more of a band-aid than anything else.

This does allow you to get more than 1 jump in whilst not moving or moving at a very low speed <100u/s however when moving at speed that actually matters then it seems to work as it should. 

I spent 10+ mins jumping in a zone and i could not break it. I also was not in any way able to use this to my advantage.


Kyli3Boi:
This is for #86

PR #267, opened 2020-11-08
Update DHooks URL

olokos:
Had to fix a few conflicts with dev branch.
I'm not sure why we've got both master and release, but they seem to be doing different things.
Possibly we keep just the release.yml and get rid of master.yml?


olokos:
From what I understand one just compiles the surftimer .smx and the other actually makes a complete package for release, so both should be kept, actually.

@Bara correct me if I'm wrong, as I know you're great with git workflows and not just that. 😃

Bara:
`release.yml` is for the master branch and `master.yml` for dev branch/compile testing, yeah `master.yml` is probably the wrong name should be renamed to `test.yml` or `dev.yml` or so.

olokos:
Done!

olokos:
Already fixed

PR #275, opened 2021-03-12
Add dowgrade file for newer checkpoints (and other changes)

Bara:
If you get errors with the drop column query, delete them from the file.

Tested by @freakexeuLow and me

**Make backup before you use this. Usage at your own risk!**


olokos (approved):
If dropping `ck_announcements` and the rest isn't causing any actual useful data loss, then this looks great and very useful for those willing to test

Bara:
I revert the changes from [upgrade 5 and 6](https://github.com/surftimer/Surftimer-Official/blob/dev/addons/sourcemod/scripting/surftimer/upgrades.sp#L77-L83). So it should be fine to dropping steamid column from ck_announcements.

Also I'll cherry-pick it to the dev-Branch and switch the target branch here from master to dev, which makes more sense.

PR #283, opened 2021-08-04
fix hookzone issue

dPexxIAM:
fixes https://github.com/surftimer/Surftimer-Official/issues/108

ty https://github.com/t5mat


t5mat:
Cool :)

Since we are not using `g_mapZones[id].TargetName` anymore, we should probably remove it from the DB in the future

Bara (approved):
Seems okay, maybe what @t5mat wrote if TargetName is never used now to remove it.

PR #284, opened 2021-08-18
Replace SQL query for style map-top to prevent duplicate rows

1zc:
Replacing the SQL query used in `db_selectStyleMapTopSurfers` to the same one used in `db_selectMapTopSurfers`. The old query brings up duplicate entries and a bunch of unnecessary rows by not searching for the right info. 

Old query vs New query: _(Sample used was `surf_tendies` on low-gravity with a total of 30 completions.)_
image

Old query result: _(First 15 rows)_
image

New query result: _(First 15 rows)_
image


Bara (approved):
See no issues here, but would nice when someone, who use this, can confirm this "fix".

1zc:
If someone else would like to test that would be awesome, but this is my testing from in-game low-gravity /mtop for `surf_tendies` on GFL.

Old:
image


New:
image

DaLLuZZ:
I just want to say that such a problem with mtop and srcp actually existed in the plugin. But this solution does not fully satisfy me. Why didn't you fix the query for srcp but only for mtop?
https://github.com/surftimer/Surftimer-Official/blob/3a492cc7ed4aa2ef23edcbb277c59969f721d9e2/addons/sourcemod/scripting/surftimer/sql.sp#L7318-L7321
I can also recommend you taking a look at the callback function. I think that you should delete some part of the code there. Maybe not in this PR, but keep in mind...
https://github.com/surftimer/Surftimer-Official/blob/3a492cc7ed4aa2ef23edcbb277c59969f721d9e2/addons/sourcemod/scripting/surftimer/sql.sp#L2116-L2122
I think you don't need to check if you have duplicate strings in result, because they are not possible due to primary key
https://github.com/surftimer/Surftimer-Official/blob/3a492cc7ed4aa2ef23edcbb277c59969f721d9e2/addons/sourcemod/scripting/surftimer/sql.sp#L148
Exactly the same should be done in `sql_selectStageStyleTopSurfersCallback`
https://github.com/surftimer/Surftimer-Official/blob/3a492cc7ed4aa2ef23edcbb277c59969f721d9e2/addons/sourcemod/scripting/surftimer/sql.sp#L7331

Bara (commented):
I looked into the code, how about merging this query into `sql_selectTopSurfers`?

Bara (approved):

Bara:
Looks good for me, but test with approval would be better. 👌🏻

PR #285, opened 2021-08-18
Add LG/SM cTime thresholds for prehop prevention

1zc:
I figured out a way to extend the existing pre-hop prevention to accommodate the slower styles (Low-Gravity and Slow-Motion) by adding cTime thresholds that meet the style's requirements when checking for whether a player is currently in a bhop.

I initially tried re-working the system in agreement with sneaK's comment that the current system is pretty bad, but I ended up just giving up and trying to extend the current system. Until we get a full re-work, this should prevent pre-hopping in both Low-Gravity and Slow-Motion.

cTime Measurements: 
image

### DIFF
--- a/addons/sourcemod/scripting/surftimer/hooks.sp
+++ b/addons/sourcemod/scripting/surftimer/hooks.sp
@@ -1374,7 +1374,10 @@ public Action Event_PlayerJump(Handle event, char[] name, bool dontBroadcast)
 						g_bJumpedInZone[client] = true;
 						g_bResetOneJump[client] = true;
 						g_fJumpedInZoneTime[client] = GetGameTime();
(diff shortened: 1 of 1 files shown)


dPexxIAM (commented):
seems fine, thanks! although i think you forgot about https://github.com/surftimer/Surftimer-Official/blob/b6a71e9ebde21cb865464c932de116baec199bf6/addons/sourcemod/scripting/surftimer/hooks.sp#L1351

1zc:
@dPexxIAM It appears I have, I'll add LG/SM thresholds for that later today

1zc:
@dPexxIAM I've updated the one-jump limiter to work with the LG and SM styles, let me know if I've missed something!

dPexxIAM (approved):

PR #292, opened 2021-09-05
Specify style = 0 when running db_selectPlayerRank

1zc:
This is a tiny modification to the queries used in the `db_selectPlayerRank` function. When players use the `/rank @#` command to check someone's rank and points in the chat, the `db_selectPlayerRank` can display style ranks and points, often leading to confusion.

A good example of this can be seen in the screenshots below. The first screenshot showcases the `/rank @1` command before this fix - the expected result should be "w", but we instead see Dini's slow motion top rank and score.
image

With this fix, issuing the same command gives us the expected results:
image


Bara (approved):
atm. I'm fine with it. I keep this open for another opinion.

PR #294, opened 2021-09-06
fix sm1.11 build

Ciallo-Ani:


addons/sourcemod/scripting/surftimer/sql.sp
```diff
@@ -6814,7 +6814,7 @@ public void db_GetStyleMapRecord_Pro(int style)
 	SQL_TQuery(g_hDb, sql_selectStyleMapRecordCallback, szQuery, style, DBPrio_Low);
 }
 
-public void sql_selectStyleMapRecordCallback(Handle owner, Handle hndl, const char[] error, int style)
+public void sql_selectStyleMapRecordCallback(Handle owner, Handle hndl, const char[] error, any style)
```
Bara, line 6817: This change [isn't required anymore](https://github.com/alliedmodders/sourcepawn/issues/613#issuecomment-903326279) and will revert within the next sourcepawn update, but both ways shouldn't hurt and we don't know how long it takes for the sourcepawn update - alternative we can [set the sp compiler](https://github.com/surftimer/Surftimer-Official/blob/bdd88bbb367b99accd160c63c2ff5d108b440e30/.github/workflows/master.yml#L13) to `1.11.6735`.

addons/sourcemod/scripting/surftimer/sql.sp
```diff
@@ -6854,7 +6854,7 @@ public void db_viewStyleMapRankCount(int style)
 	SQL_TQuery(g_hDb, sql_selectStylePlayerCountCallback, szQuery, style, DBPrio_Low);
 }
 
-public void sql_selectStylePlayerCountCallback(Handle owner, Handle hndl, const char[] error, int style)
+public void sql_selectStylePlayerCountCallback(Handle owner, Handle hndl, const char[] error, any style)
```
Bara, line 6857: This change [isn't required anymore](https://github.com/alliedmodders/sourcepawn/issues/613#issuecomment-903326279) and will revert within the next sourcepawn update, but both ways shouldn't hurt and we don't know how long it takes for the sourcepawn update - alternative we can [set the sp compiler](https://github.com/surftimer/Surftimer-Official/blob/bdd88bbb367b99accd160c63c2ff5d108b440e30/.github/workflows/master.yml#L13) to `1.11.6735`.

addons/sourcemod/scripting/SurfTimer.sp
```diff
@@ -2897,7 +2897,7 @@ public APLRes AskPluginLoad2(Handle myself, bool late, char[] error, int err_max
 
 /*======  End of Natives  ======*/
 
-public Action ItemFoundMsg(UserMsg msg_id, Handle pb, const players[], any playersNum, any reliable, any init)
+public Action ItemFoundMsg(UserMsg msg_id, BfRead msg, const int[] players, int playersNum, bool reliable, bool init)
```
Bara, line 2900: CSGO use Protobuf, so in this case we should use [the second Typeset](https://sourcemod.dev/#/usermessages/typeset.MsgHook).

Bara (changes requested):

Bara (approved):

PR #298, opened 2021-09-09
another fix for Trigger OnEndTouch issue that keeping velocity when using teleport command

Ciallo-Ani:
<del>not 100% fix if someone set their startzone inside a map trigger OnEndTouch boost zone</del> 
thanks dpexx's idea

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sm_version: [ "1.10", "1.11" ]
+        sm_version: [ "1.10", "1.11.6735" ]
 
 
     steps:
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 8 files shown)


dPexxIAM (commented):
it seems like this fix isn't doing anything different that what i PR'd, other that changing variable names and calling them slightly different. unless i'm wrong..

Any reason why you set g_bTeleByCommand[client] = true; in each commands function rather than just calling it inside TeleportClient?

Ciallo-Ani:
just imaging ur standing on a big ground that full of endtouch trigger, and u cannot back to the start zone by command directly

i am not sure whether `TeleportClient` function will be only used command or not in the future

dPexxIAM:
Did you test your code? When I implemented it, it seemed to just nuke the output of the trigger entirely.




> just imaging ur standing on a big ground that full of endtouch trigger, and u cannot back to the start zone by command directly
> 
> i am not sure whether `TeleportClient` function will be only used by command or not in the future

This shouldn't cause that many issues but i can agree it isn't ideal.

You should use TeleportClient instead.

dPexxIAM (approved):
looks good, thanks!!

Issue #98, opened 2020-05-05
[Enhancement] Improve bots movetype by changing it to noclip


olokos:
**Describe the bug**
Bots might not be perfectly smooth, as reported by some people

**Expected behavior**
Bots should be as smooth as an actual player

**Additional context**
@ghost has submitted a PR that changed the movetype to noclip
#20 
We need to reevalute smoothness of bots movements, especially with the upcoming 286 release.
I haven't noticed any issues myself, but it's worth trying out if bot movetype noclip is an improvement or not.

deadw1nter:
I think the only way to make bots smooth is rewriting the whole replay. Because the code is kind of messy and overcomplicated.

qawery-just-sad:
tbh this whole plugin needs a rewrite

olokos:
I suppose setting noclip didn't fix the issue and a comment with this idea is already in the code, so closing this.

PR #310, opened 2021-09-21
Update and fix replay

Ciallo-Ani:
Sometimes the bots bug when a player cuts a trigger/wall too close, and it has been bothering us for a long time:rage:
Also, this problem can be traced back to the time since cksurf use **velocity** and **angle** to store our replay
(pretty worse legacy:joy:)

some like #20 #98 point it out directly but no one solve.

This PR cost me a half day, and here's what i'm mainly doing

- Replace the function with methodmap
- **Support the old replay** version(0x01) while using the new version(0x02)
- Write and load new replay based on **position** and **angle**:blush::blush::blush:
- Delete useless(unused) sentences and improve something

**I haven't done enough testing yet**, may have bugs
but hope that there's something change since this pr :persevere::persevere::persevere:.

### DIFF
--- a/addons/sourcemod/scripting/SurfTimer.sp
+++ b/addons/sourcemod/scripting/SurfTimer.sp
@@ -112,7 +112,7 @@
 
 // Replay Definitions
 #define BM_MAGIC 0xBAADF00D
-#define BINARY_FORMAT_VERSION 0x01
+#define BINARY_FORMAT_VERSION 0x02
 #define ADDITIONAL_FIELD_TELEPORTED_ORIGIN (1<<0)
 #define ADDITIONAL_FIELD_TELEPORTED_ANGLES (1<<1)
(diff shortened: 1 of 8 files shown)


dPexxIAM (commented):
this is a big change, and a lot of communities will be happy about this! thank you.

code seems good, but like you said, testing needed

1zc:
Did some testing:

- Tested with some older broken replays we had in GFL, and they now work as expected with your PR! 
- Old replays work fine with the new version, didn't encounter any potential compatibility issues,
- In an effort to break it, I also tested writing ridiculous replays of me just purposely clipping platforms as close as I could. Moved in stupid random ways & reset several times but could not break it,

Great work! Excited to see these issues finally be a thing of the past.

dPexxIAM (approved):
epic

Issue #319, opened 2021-09-28
[BUG] Changing Client Noclip Speed Causes DispatchAsyncEvent Spam


CustomStats:
**Describe the bug**
Changing the `!noclipspeed` value to anything other than the default `sv_noclipspeed` value causes the console to be spammed with DispatchAsyncEvent backlog errors, resulting in server-wide FPS lag that will not stop until the value is set back to default.

**Expected behavior**
Should have no effect on anything server-side.

**Screenshots**
https://streamable.com/vgiiyx

**Plugin List**
SurfTimer 1.0.1

**Additional context**
This only began happening after the recent operation and we've seen this same thing in other plugins that changed client variables through SDKHook_Think. Server is running on linux.

PR #320, opened 2021-09-28
fix change noclipspeed causes DispatchAsyncEvent spam

Ciallo-Ani:
#319


Bara:
Is this tested and confirmed to work/to fix the issue?

Ciallo-Ani:
> Is this tested and confirmed to work/to fix the issue?

tested

Bara (approved):

PR #321, opened 2021-09-28
Disable practice mode when player touches a map/bonus start zone.

1zc:
Currently, the timer disables Practice Mode for a player only when they exit the start zone. This has left room for an exploit in the current release that allows players to bypass limitations placed in the start-zone as a result of still being in Practice Mode after failing. 

Since `g_bPracticeMode[client]` is still true when a player is in the start-zone after failing, they can bypass the prespeed limiter and prehop infinitely to give themselves a massive advantage. (i.e, players can enter practice mode, fail by hitting a reset trigger, and then prehop infinitely in the start zone.)

This pull request turns off Practice Mode for a player when they enter a start zone rather than when they exit one. This prevents the above exploit from being possible.


dPexxIAM (approved):
ty!

PR #325, opened 2021-09-29
fix replay hud speed

Ciallo-Ani:
**** : when i set a record as a test the units of the run on the bot are like 200 more units than what i had on 285 surftiemr they match fine its weird


dPexxIAM:
Thanks, can you also test style replays that modify movement speed with fLaggedMovement? I suspect those units will be incorrect aswell.

Ciallo-Ani:
> Thanks, can you also test style replays that modify movement speed with fLaggedMovement? I suspect those units will be incorrect aswell.

FF and LG is also fixed (see discord)

dPexxIAM (approved):
thanks!

PR #327, opened 2021-09-29
fix hud cant be disabled when player's dead

Ciallo-Ani:
player can not disable their hud when dead, even if they already set their timer's sidehud or centre hud to **[OFF]**


Ciallo-Ani:
tested

dPexxIAM (approved):
ty

PR #328, opened 2021-09-30
Resolve #114 - Update LimitMaxSpeed in hooks to use the correct MaxSpeed value

### LINKED ISSUE #114: [Suggestion] Speed Limit Zone
**Is your feature request related to a problem? Please describe.**
Superboosting. The source engine sucks and allows you to activate triggers more often than you should.

**Describe the solution you'd like**
Since there is really no clear-cut way to eliminate superboosting, so a zone with a determined allowed max velocity would be ideal for server owners/zoners. This would invalidate map/bonus/stage timers if passed through at a higher than acceptable speed.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

1zc:
To resolve #114,

This PR implements what @timyc mentioned in the above issue. It allows zoners to use Max Speed zones and takes into account the actual speed set. 

From my testing, this also seems to make MaxSpeed zones work when they didn't before.
Here's a quick video of me testing it: [Streamable](https://streamable.com/c02cc5)

### DIFF
--- a/addons/sourcemod/scripting/surftimer/hooks.sp
+++ b/addons/sourcemod/scripting/surftimer/hooks.sp
@@ -899,7 +899,7 @@ public Action OnPlayerRunCmd(int client, int &buttons, int &impulse, float vel[3
 	}
 	else if (g_bInMaxSpeed[client])
 	{
-		LimitMaxSpeed(client, 2500.0);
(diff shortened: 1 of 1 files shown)


dPexxIAM (approved):
LOL...

thanks

PR #332, opened 2021-10-06
fix Array index out-of-bounds error when Command_ToStage is used on Linear maps, fix command_tostage !s 3 bug

dPexxIAM:
should fix 

Exception reported: Array index out-of-bounds (index 45, limit 37)
Blaming: SurfTimer.smx
Call stack trace:
  [1] Line 220, surftimer/misc.sp::teleportClient
  [2] Line 1289, surftimer/commands.sp::Command_ToStage


dPexxIAM:
> Translations? 🤔

fixed

PR #342, opened 2021-10-13
Update misc.sp

Nubbeezez:
Edited BotQuota so that it doesnt fuck fps for client and fixes VAR spikes on server :)
Atleast made my personal server stop eating my fps
newb to sourcepawn/coding in total but trying to learn 👍 
Please give me criticism

### DIFF
--- a/addons/sourcemod/scripting/surftimer/misc.sp
+++ b/addons/sourcemod/scripting/surftimer/misc.sp
@@ -1,10 +1,15 @@
-public void SetBotQuota()
+void setBotQuota()
 {
 	// Get bot_quota value
 	ConVar hBotQuota = FindConVar("bot_quota");
 
+	int flags = GetConVarFlags(hBotQuota);
+	flags &= ~FCVAR_REPLICATED;
+	SetConVarFlags(hBotQuota, flags);
+
 	// Initialize
 	SetConVarInt(hBotQuota, 0, false, false);
+
 
 	// Check how many bots are needed
 	int count = 0;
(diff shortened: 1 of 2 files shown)


addons/sourcemod/scripting/surftimer/misc.sp
```diff
@@ -1,11 +1,16 @@
-public void SetBotQuota()
+void setBotQuota()
```
DaLLuZZ: Don't change the name of the function, please

DaLLuZZ (changes requested):

addons/sourcemod/scripting/surftimer/misc.sp
Nubbeezez: I did just change it back :) sorry

DaLLuZZ:
```cpp
public void SetBotQuota()
```
instead of
```cpp
public void setBotQuota()
```
please

Nubbeezez:
So if im understanding this correctly ? its the maps fault that its getting spammed after the latest riptide update ?

Bara:
> removing FCVAR_REPLICATED is not the best way to solve the problem

Yes, but instead of updating every map (whenever the mapper is still active) it's a worth to add this workaround. Just depends on if it's a map issue.

1zc (approved):
As concluded on Discord, there is no real need for bot_quota to be replicated to the client and there is no harm in preventing it from happening. LGTM!

PR #351, opened 2021-11-10
LimitMaxSpeed only when zoneID is not below zero, attempt to resolve #350

### LINKED ISSUE #350: [BUG] Exception reported: Array index out-of-bounds (index -1, limit 128)
**Describe the bug**
Before this topic goes down, I wanted to post it on Github. When exactly the problem comes I can't say, but I have recent logs of it that have over 300+ mb and hope that someday someone nice and social might fix this problem.

```
L 11/03/2021 - 00:00:00: SourceMod error session started
L 11/03/2021 - 00:00:00: Info (map "surf_paradigm") (file "/server/csgo/csgo/addons/sourcemod/logs/errors_20211103.log")
L 11/03/2021 - 00:00:00: [SM] Exception reported: Array index out-of-bounds (index -1, limit 128)
L 11/03/2021 - 00:00:00: [SM] Blaming: SurfTimer.smx
L 11/03/2021 - 00:00:00: [SM] Call stack trace:
L 11/03/2021 - 00:00:00: [SM]   [1] Line 902, surftimer/hooks.sp::OnPlayerRunCmd
L 11/03/2021 - 00:00:00: [SM] Exception reported: Array index out-of-bounds (index -1, limit 128)
```

**Expected behavior**
That the SurfTimer plugin does not spam the server with many logs :P

**Plugin List**
SurfTimer -->Master<-- Version

1zc:
I've talked briefly about this in #350 - this pull request adds a simple check to ensure that the zoneID is valid before attempting to limit speed when the player is in a MaxSpeed zone.

### DIFF
--- a/addons/sourcemod/scripting/surftimer/hooks.sp
+++ b/addons/sourcemod/scripting/surftimer/hooks.sp
@@ -899,7 +899,8 @@ public Action OnPlayerRunCmd(int client, int &buttons, int &impulse, float vel[3
 	}
 	else if (g_bInMaxSpeed[client])
 	{
-		LimitMaxSpeed(client, g_mapZones[g_iClientInZone[client][3]].PreSpeed);
+		if (g_iClientInZone[client][3] >= 0)
+			LimitMaxSpeed(client, g_mapZones[g_iClientInZone[client][3]].PreSpeed);
 	}
 
 	/*------ Styles ------*/


DaLLuZZ:
I am not sure, but as far as I know MaxSpeed zone will not work after this

DaLLuZZ:
check this
https://github.com/surftimer/Surftimer-Official/blob/master/addons/sourcemod/scripting/surftimer/surfzones.sp#L224-#L249

DaLLuZZ:
check this
https://github.com/surftimer/Surftimer-Official/blob/master/addons/sourcemod/scripting/surftimer/surfzones.sp#L224-#L249

1zc:
Yep, you seem to be right, haven't tested it though - however, for now, I think its best we first get rid of extreme spamming of everyone's log files and focus on fixing the feature's entire functionality at a later time.

I did test that this works for MaxSpeed zones placed outside of any other zone, so it should still function to an extent. I'll try figuring something out when I get the time, but perhaps for another PR. 

For now, the focus of this is to directly resolve #350

addons/sourcemod/scripting/surftimer/hooks.sp
```diff
@@ -899,7 +899,8 @@ public Action OnPlayerRunCmd(int client, int &buttons, int &impulse, float vel[3
 	}
 	else if (g_bInMaxSpeed[client])
 	{
```
olokos, line 901:
```
if(client){
return;
}

```
What about something like this? Possibly even as the first condition of this entire method?

DaLLuZZ: useless condition

1zc:
While this is not a full-fledged fix for MaxSpeed zones not functioning as intended - for now, merging this so that the plugin doesn't spam server logs.

Bara (approved):

PR #354, opened 2021-11-17
New forward and separate Discord module

Sarrus1:
Hello,

This PR adds a new forward to `surftimer.inc` which allows other plugins to receive informations on when a player sets a new record.
This is in an effort to separate the Discord module from the main plugin, as others have requested previously.
I've been working on a separate plugin already, you can find it [here](https://github.com/Sarrus1/SurfTimer-discord).

Here is an example:

Image

I've also added an enum in this (eca0ba3b93df997cc92477edd71babc0770ad17e) commit for clarity.

✅This has been tested on an empty server, with bonus and regular records.


addons/sourcemod/scripting/SurfTimer.sp
```diff
@@ -2718,6 +2724,7 @@ public void OnPluginStart()
 	g_MapCheckpointForward = CreateGlobalForward("surftimer_OnCheckpoint", ET_Event, Param_Cell, Param_Float, Param_String, Param_Float, Param_String, Param_Float, Param_String);
 	g_BonusFinishForward = CreateGlobalForward("surftimer_OnBonusFinished", ET_Event, Param_Cell, Param_Float, Param_String, Param_Cell, Param_Cell, Param_Cell);
 	g_PracticeFinishForward = CreateGlobalForward("surftimer_OnPracticeFinished", ET_Event, Param_Cell, Param_Float, Param_String);
+	g_NewRecordForward = CreateGlobalForward("surftimer_OnNewRecord", ET_Event, Param_Cell, Param_Cell, Param_String, Param_String, Param_Cell);
```
Bara, line 2727: When you change the global vars. from Handle to GlobalForward, you should change this to `new GloobalForward` too

Bara (approved):

PR #355, opened 2021-11-20
Replace ColorVariables with ColorLib

Bara:
More details in my older PR #66


Bara:
Seems like setup-sp doesn't work correctly since latest sourcepawn compiler update.

1zc (approved):
Good stuff!

PR #356, opened 2021-11-24
Replace SQL_GetRowCount with SQL_FetchInt in SQL_ViewAllRecordsCallba…

bctse:
…ck2 in sql.sp

This should allow the ranks to show up correctly rather than showing everyone as rank 1 when searching their completed maps.


1zc (approved):
Good catch, LGTM. Thank you!

PR #367, opened 2022-01-16
Feat: Add new `OnNewWRCP` forward

Sarrus1:
This PR adds a new forward which gets called when a new WRCP is set.

```c
/**
 * Called when a client sets a new WRCP
 *
 * @param client			The client's ID.
 * @param style				Style index.
 * @param time				Time set by the player.
 * @param timeDif			Time difference with the former record.
 * @param stage			        The id of the stage of the WRCP.
 */
forward void surftimer_OnNewWRCP(int client, int style, char[] time, char[] timeDif, int stage);
```


dPexxIAM (approved):
lgtm

PR #369, opened 2022-01-17
Fix: `surftimer_OnNewWRCP` forwards formatted time

Sarrus1:


Sarrus1:
There seems to still be a random bug with the value of `stage`, I will fix it asap

dPexxIAM (approved):

PR #370, opened 2022-01-18
Add a new stock for clarity and remove an old stock

Sarrus1:
This PR adds a new stock `EmitSoundToClientNoPreCache` which allows to play a sound to a client without having to precache that sound.
This takes advantage of [this CSGO quirk](https://wiki.alliedmods.net/Csgo_quirks#The_.22play.22_client_command).
It was already implemented in the plugin, the PR just makes it more declarative through a stock.

```c
/**
 * Emit a sound to a client without having to precache the sound.
 * See https://wiki.alliedmods.net/Csgo_quirks#The_.22play.22_client_command for more details.
 * 
 * @param client      The client's id.
 * @param szPath      The path of the sound with or without "play " as a suffix.
 * @param addPlay     Whether or not "play " should be append to the begining of szPath (default = true).
 */
stock void EmitSoundToClientNoPreCache(int client, const char[] szPath, bool addPlay = true)
{
	char szBuffer[256];
	if(addPlay)
	{
		Format(szBuffer, sizeof szBuffer, "play %s", szPath);
	}
	else
	{
		strcopy(szBuffer, sizeof szBuffer, szPath);
	}
	ClientCommand(client, szBuffer);
}
```

### DIFF
--- a/addons/sourcemod/scripting/surftimer/buttonpress.sp
+++ b/addons/sourcemod/scripting/surftimer/buttonpress.sp
@@ -12,7 +12,7 @@ public void CL_OnStartTimerPress(int client)
 				if (fGetGameTime - g_fErrorMessage[client] > 1.0)
 				{
 					CPrintToChat(client, "%t", "BPress1", g_szChatPrefix);
-					ClientCommand(client, "play buttons\\button10.wav");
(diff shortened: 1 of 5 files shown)


dPexxIAM (approved):
thanks!

PR #372, opened 2022-01-21
Fix #371

### LINKED ISSUE #371: [BUG] `surftimer_OnNewRecord` is called multiple time on new records
**Describe the bug**
The forward `surftimer_OnNewRecord` gets called multiple time on new records. It is called exactly the number of players on the server. So if there are two people, it will get called 2 times, if there are 3, three times etc.

**Expected behavior**
It should only get called once.

**Additional context**
This happens in v1.0.5, and did not happen before.

Sarrus1:


tslashd:
Tested with 2 players online. 
Only 1 message was sent to Discord upon a record, regardless if one of the players was spectator or their timer was running

dPexxIAM (approved):
THNX!

PR #373, opened 2022-01-22
Prestrafe comparison

Topvennie:
New feature which shows the difference between your prestrafe speed and the map record.

On a linear map or the first stage of a staged map it compares with the prestrafe speed of the map record.
For the other stages it compares with the stage record.
Bonus prestrafes are compared bonus records.


dPexxIAM (approved):
LGTM! Thanks for this!!

Bara:
And from where do we get the "ghost tables"? :)

dPexxIAM:
As bara said, this will require an update script for the DB schema before we can push a new release with these changes.

Topvennie:
Not sure what you mean with ghost tables
But if you mean the prestrafe data itself then if there's no saved data aka the prestrafe speed = 0 then it doesn't show the difference and just prints "start prestrafe: 340 u/s"

It also starts saving the prestrafe data.

Worked for me with the current Db schema

Bara:
@Topvennie you seems to use the OldDev branch or another fork with the db changes. Offical the ck_bonus has the following columns: steamid, name, mapname, runtime, zonegroup and style

Bara:
You could look into the "CREATE TABLE" queries >[here](https://github.com/surftimer/Surftimer-Official/blob/master/addons/sourcemod/scripting/surftimer/db/queries.sp)<.

Topvennie:
Ah my bad...
I'll add a DB update script asap

PR #376, opened 2022-01-24
Hints

Topvennie:
This PR adds helpfull hints for new players when it comes to commands. Also allows server owners to easily change 
- The hints themselves
- The time between two hints
- The order in which they are displayed

The hints.txt file supports {colours} and a maximum size of 265 / hint


dPexxIAM (changes requested):
Good idea, thanks! Although some of the hints you wrote are already being printed to chat.

https://github.com/surftimer/Surftimer-Official/blob/master/addons/sourcemod/translations/surftimer.phrases.txt#L53

You should remove any which are redundant to avoid spam before we merge this.

Topvennie:
I removed the timers which print them but forgot to remove them from the .phrases.txt file

Anyway they're gone now

addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -2251,6 +2265,17 @@ void PrespeedText(int client, bool menu = false)
 		MiscellaneousOptions(client);
 }
 
+void HintsText(int client, bool menu = false)
+{
+	if (g_bAllowHints[client])
```
Bara, line 2270: `g_bAllowHints[client] = !g_bAllowHints[client];`
in a smaller easier version instead of the if/else

Topvennie: I orginally just copied the prestrafe functions. Changed that one as well

dPexxIAM (approved):

PR #379, opened 2022-01-25
Center HUD Fixes #1

### LINKED ISSUE #1: [Suggestion] Server record message improvements
**Is your feature request related to a problem? Please describe**.
No

**Describe the solution you'd like**
Have the timer show the difference between old SR and new SR, when the SR is beaten. This is currently done with the bonus record, but not the map itself.

_Some ideas for formatting:_
Username finished in X beating the SR record by -00:00:01. Improving their best time by Y
Username beat the map record!

Username finished in X (SR -00:00:01). Improving their best time by Y
Username beat the map record!

Usernamr finished in X. Improving their best time by Y.
Username beat the map record! (SR -00:00:01)

**Describe alternatives you've considered**

**Additional context**
Here's a picture of the bonus server record message, where it actually compares the new record with the previous server record (instead of the client's previous own PB)

https://user-images.githubusercontent.com/57944222/79012267-38ea9600-7b66-11ea-9379-ad3acd2791b3.png

And here's how it looks when a player gets the map record:

https://user-images.githubusercontent.com/57944222/79012932-d5616800-7b67-11ea-8791-5c30f2fc7a91.png

tslashd:
There is an issue with the **HUD** when you are passing through a checkpoint and times switch to **+** or **-** comparisons, the last digit of _**Speed:**_ would disappear along with _**u/s**_.
old 2

After some testing I came to the conclusion that this is due to character limit in the **HUD**, so I went on a quest to save some characters. By removing the spaces between _**Stage: %i / i%**_ and _**Rank: %i / i%**_ that issue was gone but it needed a little bit more as it was still cutting off last digit and I removed the space whenever **+** or **-** comparisons appear. But even after all of that it was still cutting off _**u/s**_, so I removed it as it flashes on and off when passing through stages/checkpoints which is distracting and altogether not cool (sue me).
new 2

Removing those spaces caused some issues with spacing between modules in the **HUD**
, so I edited those a little bit as _**SR: XX:XX:XX**_ was always too close to _**00:00:00**_ (ocd soz).
**_Before edit:_**
old 1
**_After edit:_**
new 1

At the moment **HUD** looks great in normal mode and normal practice mode **_(styled HUD is still all over the place, will deal with those in another PR)_**
_**Before edit:**_
old 3

_**After edit:**_
new 3

_P.S: Styled HUD was **fricked** even before these edits_

### DIFF
--- a/addons/sourcemod/scripting/surftimer/misc.sp
+++ b/addons/sourcemod/scripting/surftimer/misc.sp
@@ -1,3 +1,5 @@
+
+
 public void SetBotQuota()
 {
 	// Get bot_quota value
@@ -9,7 +11,6 @@ public void SetBotQuota()
 
 	// Initialize
 	SetConVarInt(hBotQuota, 0, false, false);
-
 
 	// Check how many bots are needed
 	int count = 0;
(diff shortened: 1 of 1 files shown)


dPexxIAM (approved):
Thank you!

Issue #375, opened 2022-01-23
[BUG] No sound played and no time difference shown on map recs


tslashd:
**Describe the bug**
Upon beating a record of a map, no sound is being played and also time difference between previous SR is not shown.

**Expected behavior**
Play a sound and show difference in time between current SR and previous one

**Screenshots**
None

**Plugin List**
SurfTimer 1.0.5 release, including #372 _(which fixes multiple surftimer_OnNewRecord calls)_

dPexxIAM:
should be fixed with https://github.com/surftimer/Surftimer-Official/pull/381

PR #381, opened 2022-01-27
No sounds bug

Topvennie:
Fixes the sound bug #375 

Haven't actually tested it but it should work.

### DIFF
--- a/addons/sourcemod/scripting/SurfTimer.sp
+++ b/addons/sourcemod/scripting/SurfTimer.sp
@@ -859,6 +859,9 @@ char g_szTimeDifference[MAXPLAYERS + 1][32];
 // Record map time in seconds
 float g_fRecordMapTime;
 
+// Old map record time in seconds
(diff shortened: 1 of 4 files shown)


Topvennie:
Added a wr difference message after "{1} {yellow}{2} {default}beat the map record!"

dPexxIAM (approved):
ty!

tslashd:
Does this fix the issue with styled map and bonus SR's not being forwarded correctly? @Topvennie

dPexxIAM:
> Does this fix the issue with styled map and bonus SR's not being forwarded correctly? @Topvennie

No. This was never mentioned anywhere on Git. Make another issue.

tslashd:
> > Does this fix the issue with styled map and bonus SR's not being forwarded correctly? @Topvennie
> 
> No. This was never mentioned anywhere on Git. Make another issue.
ty #382

PR #386, opened 2022-02-03
Make times accurate

### LINKED ISSUE #282: [BUG] Timer is inaccurate
**Describe the bug**
Copied from discussion in the discord.

These 2 issues make the timer inaccurate,

1. Run time is calculated using GetGameTime(). This is bad because time should be based on playercmds and not server frames, since multiple playercmds may be processed in a single frame. Instead code should count PostThink()s or some shit and use this as time, for example - https://github.com/rumourA/End-Touch-Fix/blob/72e7dc745371e91853b19c7e0980ae8953fb9e62/scripting/BareTimer.sp#L96

https://github.com/surftimer/Surftimer-Official/blob/b6a71e9ebde21cb865464c932de116baec199bf6/addons/sourcemod/scripting/surftimer/buttonpress.sp#L189

2. (I guess this one becomes relevant after fixing 1) The engine calls EndTouch()s at the end of each server frame, not after each ProcessMovement. So if multiple playercmds are processed in 1 frame, EndTouch might be delayed by a few "player ticks". Fix is fairly simple - https://github.com/rumourA/End-Touch-Fix/blob/72e7dc745371e91853b19c7e0980ae8953fb9e62/scripting/EndTouchFix.sp#L73

t5mat:
#### Summary of changes:

- **Calculate times using client ticks instead of GetGameTime()** - Time should be calculated based on the number of executed playercmds and not server frames, since multiple playercmds (or none) may be processed in a single server frame. [Also, GetGameTime() is weird and returns a lag compensated value (?)](https://hlcoders.valvesoftware.narkive.com/IG3IBv9m/gpglobals-curtime). In any case, this change makes times independent of lag.

- **Make checkpoints accuracy higher than 0.1s** - Before, checkpoint times were only updated in 0.1s intervals (??)

- **Make EndTouchFix a requirement** - The engine calls EndTouch()s at the end of each server frame, not after each ProcessMovement. So if multiple playercmds are processed in 1 frame, trigger EndTouch calls might be delayed by a few client ticks. EndTouchFix fixes this issue and is required for having accurate times.

Resolves #282.

**Before/after video demonstrating the fix (same saveloc is restored but time is different after each completion):**

https://user-images.githubusercontent.com/16616463/152370121-c59112a8-9c9f-494b-9aea-f3d1dee450c0.mp4

[**Bonus exploit that is hopefully prevented now**](https://www.youtube.com/watch?v=p1AtpEcMnDk)

### I originally implemented this for [GoFree](https://stats.go-free.info) & @dPexxIAM. We've both decided to make this change public. Hopefully, this will make other people follow suit with sharing their bug fixes and contribute to SurfTimer :relaxed:

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,10 +48,10 @@ Yes, but not any more than other projects currently and for past years.
 * [AutoExecConfig](https://github.com/Impact123/AutoExecConfig)
 
 **SourceMod Plugins**
+* [End-Touch-Fix](https://github.com/rumourA/End-Touch-Fix) - Checks EntityUntouch on PostThink instead of server frames
 * *(recommended)* [Movement Unlocker](https://forums.alliedmods.net/showthread.php?t=255298) - Enables ground sliding
 * *(recommended)* [MomSurfFix](https://github.com/GAMMACASE/MomSurfFix) - Fixes ramp glitches
 * *(recommended)* [RNGFix](https://github.com/jason-e/rngfix) - Fixes a bunch of engine physics "bugs"
-* *(recommended)* [End-Touch-Fix](https://github.com/rumourA/End-Touch-Fix) - Checks EntityUntouch on PostThink instead of server frames
 * *(recommended)* [HeadBugFix](https://github.com/GAMMACASE/HeadBugFix) - Fixes the head boundary box poping up when you start ducking
 * *(recommended)* [PushFixDE](https://github.com/GAMMACASE/PushFixDE) - Fixes client prediction errors in push triggers
 * *(recommended)* [crouchboostfix](https://github.com/t5mat/crouchboostfix) - Prevents crouchboosting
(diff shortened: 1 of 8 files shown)


Bara:
You could use `LibraryExists("endtouchfix");` to check if [EndTouchFix](https://github.com/rumourA/End-Touch-Fix/blob/master/scripting/EndTouchFix.sp#L57) is loaded.

Example snippet:
```sp
public void OnAllPluginsLoaded()
{
    if (!LibraryExists("endtouchfix"))
    {
        SetFailState("Plugin \"End Touch Fix\" not found!");
        return;
    }
}
```

t5mat:
> You could use `LibraryExists("endtouchfix");` to check if [EndTouchFix](https://github.com/rumourA/End-Touch-Fix/blob/master/scripting/EndTouchFix.sp#L57) is loaded.
> 
> Example snippet:
> 
> ```sourcepawn
> public void OnAllPluginsLoaded()
> {
>     if (!LibraryExists("endtouchfix"))
>     {
>         SetFailState("Plugin \"End Touch Fix\" not found!");
>         return;
>     }
> }
> ```

sure

dPexxIAM (approved):
EPIC!

tslashd:
LOVE YOU <3

PR #394, opened 2022-02-15
Fix custom titles being truncated at 66 characters

Jay2k1:
the `sizeof` returned 66 (`MAXPLAYERS + 1`, the first field) instead of the desired 1024 (the second field) of `g_szCustomTitleColoured` which is defined like this:
https://github.com/surftimer/Surftimer-Official/blob/584571d07ff3e8b0f6b70bdd7a8a851a3ebae626/addons/sourcemod/scripting/SurfTimer.sp#L420

If someone had a very long custom title due to using a lot of colors, for example:
`{darkred}R{lightred}A{red}I{orange}N{yellow}B{lightgreen}O{green}W{blue}S{darkblue}Y{purple}A{pink}Y`
it got cut off at 66 characters due to this. Sometimes it even led to the nickname not being displayed in chat.


1zc (approved):
LGTM! Good stuff 😄

PR #395, opened 2022-02-17
Limit TeleportClient to every 2 ticks

dPexxIAM:
Without this change, people can abuse boosters by double binding teleport commands. Very easy to abuse, especially when used in conjunction with sm_startpos


dPexxIAM:
A more revised fix will be merged by @Jay2k1

PR #398, opened 2022-02-19
Stage 2 bug refix

DaLLuZZ:
https://discord.com/channels/366959507764674560/379572504542445568/414184575879610368
https://github.com/surftimer/Surftimer-Official/commit/d7892b7aadbbde5a4d8f6802471f7be5791c12a7

P.S. some of this code is useless
https://github.com/surftimer/Surftimer-Official/blob/7a59681913f04adb2e43e91fd55c5d355cc167a4/addons/sourcemod/scripting/surftimer/buttonpress.sp#L838-L842

At the end of the last stage, it increases by one (to `g_TotalStages + 1`), but immediately after that it returns to the total number of stages on map (to `g_TotalStages`)
https://github.com/surftimer/Surftimer-Official/blob/7a59681913f04adb2e43e91fd55c5d355cc167a4/addons/sourcemod/scripting/surftimer/buttonpress.sp#L844-L847

However, due to the fact that `g_bWrcpEndZone[client]` is not reset after `client`'s connecting, sometimes when you pass the first stage, you will be credited with passing the second one.

So should this `g_bWrcpEndZone` be completely removed?

### DIFF
--- a/addons/sourcemod/scripting/SurfTimer.sp
+++ b/addons/sourcemod/scripting/SurfTimer.sp
@@ -1112,9 +1112,6 @@ char g_BlockedChatText[256][256];
 // Last time an overlay was displayed
 float g_fLastOverlay[MAXPLAYERS + 1];
 
-// Stage 2 Bug Fixer
-bool g_wrcpStage2Fix[MAXPLAYERS + 1];
-
 // Is client trying to teleport inside a trigger_multiple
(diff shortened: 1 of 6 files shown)


dPexxIAM (approved):
Damn this is an old bug! Big thanks

dPexxIAM:
@DaLLuZZ No other changes here planned? Just wondering since you marked it as a draft.

PR #405, opened 2022-03-04
changed translation 1024 -> 128

Topvennie:
Max Savelocs changed to 128 but translation file never got updated.

That's the only thing ¯\_(ツ)_/¯

### DIFF
--- a/addons/sourcemod/translations/surftimer.phrases.txt
+++ b/addons/sourcemod/translations/surftimer.phrases.txt
@@ -535,7 +535,7 @@
 	"Commands8"
 	{
 		"#format"	"{1:s}"
-		"en"		"{1} How did you hit 1024 save locs?"
+		"en"		"{1} How did you hit 128 save locs?"
 	}
 	"Commands9"
 	{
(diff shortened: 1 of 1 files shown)


addons/sourcemod/translations/surftimer.phrases.txt
```diff
@@ -535,7 +535,7 @@
 	"Commands8"
 	{
 		"#format"	"{1:s}"
-		"en"		"{1} How did you hit 1024 save locs?"
+		"en"		"{1} How did you hit 128 save locs?"
```
sneak-it, line 538: I agree with the sentiment - while we are correcting the figure, a proper, less silly message would be preferable. Something along the lines of "Maximum of 128 savelocs created" perhaps.

Topvennie: I changed the message, lmk if it's alright.

sneak-it:
Looks great, thank you!

sneak-it (approved):

PR #424, opened 2022-05-18
PR info menu

shipyy:
Menu that tracks:
* Time in Zone
* Completes & Attempts
* First time it taken for 1st completion

image


dPexxIAM (commented):
tested this myself and is stable, but will wait for bara's review first

addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -359,6 +363,17 @@ public Action Command_ToggleHints(int client, int args)
 	return Plugin_Handled;
 }
 
+public Action Command_PRinfo_help(int client, int args){
+
+	if(IsValidClient(client)){
+		PrintToConsole(client, "%t", "PRinfo_help");
```
Bara, line 369: Personally I would prefer `%T`, it'll use the client language instead of the server language. `%T` requires the client index after the translation name, in this case `PrintToConsole(client, "%T", "PRinfo_help", client);`.

You can find more [here](https://wiki.alliedmods.net/Translations_(SourceMod_Scripting)#Usage_in_a_Plugin).

addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -4954,6 +4969,171 @@ public Action Command_CPR(int client, int args)
...
+			char arg1[128];
+			GetCmdArg(1, arg1, sizeof(arg1));
+
+			if (StrContains(arg1, "surf_", true) != -1)
+			{
+				db_viewPRinfoMapRank(client, g_szSteamID[client], arg1);
+			}
+			else if (StrContains(arg1, "@") != -1)
+			{
+				ReplaceString(arg1, 128, "@", "");
```
Bara, line 5000: You could use `sizeof(arg1)` here, is easier when someone wanna change the size of the string.

I mean the nearly whole surftimer code use hardcoded sizes... so I don't really care about this

addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -4954,6 +4969,171 @@ public Action Command_CPR(int client, int args)
...
+		//prinfo <bonus number>
+		case 1:{
+			char arg1[128];
+			GetCmdArg(1, arg1, sizeof(arg1));
+
+			if (StrContains(arg1, "surf_", true) != -1)
+			{
+				db_viewPRinfoMapRank(client, g_szSteamID[client], arg1);
+			}
+			else if (StrContains(arg1, "@") != -1)
```
Bara, line 4998: I think `else if (arg1[0] = '@')` could work too and it's faster. ( @ESK0 😏 ) Also for other things like the `b` check

addons/sourcemod/scripting/surftimer/sql.sp
```diff
@@ -3458,13 +3433,359 @@ public void db_deleteCheckpoints()
...
+	if (!IsValidClient(client))
+		return;
+
+	char szQuery[1024];
+	Format(szQuery, 1024, sql_selectPR, szSteamID, szMapName, 0);
+	//PrintToConsole(client,szQuery);
+	SQL_TQuery(g_hDb, db_viewPRinfoCallback, szQuery, pack, DBPrio_Low);
+}
+
+public void db_viewPRinfoCallback(Handle owner, Handle hndl, const char[] error, any pack)
```
Bara, line 3461: `any pack` -> `DataPack pack`

Bara: and don't forget to delete the datapack especially when you use `return` or the code will not calling `db_prinforuntimecallback` with `SQL_TQuery`, otherwise the handle will never deleted.

Bara: the previous comment means the whole follow up, so also in the upcoming callbacks.

Bara (changes requested):
@dPexxIAM 👌🏻 
some thigns within the first look

shipyy:
Believe i fixed everything mentioned @Bara :)

Bara:
I try to Review it today, but I'm not at home. So maybe tomorrow

addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -366,8 +366,8 @@ public Action Command_ToggleHints(int client, int args)
 public Action Command_PRinfo_help(int client, int args){
 
 	if(IsValidClient(client)){
-		PrintToConsole(client, "%t", "PRinfo_help");
-		CPrintToChat(client, "%t", "PRinfo_help_chat", g_szChatPrefix);
+		PrintToConsole(client, "%T", "PRinfo_help");
```
Bara, line 369: you should read my comment again https://github.com/surftimer/SurfTimer/pull/424#discussion_r876637798
when you're doing this, it would requires the client index behind the phrase name.
in this case it would be:
`PrintToConsole(client, "%T", "PRinfo_help", client);`

shipyy: ohhh, so thats why i was getting a index error lol, i just didnt understand how, never used it like u are suggesting ty will change it accordingly

addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -359,6 +363,17 @@ public Action Command_ToggleHints(int client, int args)
 	return Plugin_Handled;
 }
 
+public Action Command_PRinfo_help(int client, int args){
+
+	if(IsValidClient(client)){
+		PrintToConsole(client, "%T", "PRinfo_help", client);
+		CPrintToChat(client, "%T", "PRinfo_help_chat", client, g_szChatPrefix);
```
tslashd, line 370: I believe that what @Bara meant was that this does not need the 2nd `client` variable as in the translation you are only passing the `g_szChatPrefix` at the start of the message and nothing else.

Also is there a difference between `%T` and `%t` ? Just looking that all the other translation files have the `%t` format might be worth keeping to that unless something is different when using capital letter for it.

addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -4954,6 +4969,171 @@ public Action Command_CPR(int client, int args)
...
+			}
+			else if (arg1[0] == 'b')
+			{
+				ReplaceString(arg1, sizeof(arg1), "b", "");
+				int bonus_number = StringToInt(arg1);
+
+				if (0 < bonus_number < MAXZONEGROUPS)
+					db_selectPRinfoUnknown(client, g_MapRankBonus[bonus_number][client], bonus_number, g_szSteamID[client]);
+				else
+					CPrintToChat(client, "%t", "InvalidBonusID", g_szChatPrefix, bonus_number);
```
tslashd, line 5019: To touch a bit more on this [comment](https://github.com/surftimer/SurfTimer/pull/424/files#r877808261).

Here you are passing through 2 variables to the translation as that's how the translation made to be used and you'd have to pass it an `int` for it's second variable.
```
	"InvalidBonusID"
	{
		"#format"	"{1:s},{2:i}"
		"en"		"{1} Invalid bonus ID {2}"
	}
```

shipyy: from my testing when using clients laguange ("%T") i do need to pass the client index, when using("%t") that is not necessary :)

addons/sourcemod/scripting/surftimer/commands.sp
Bara: %t uses the server language (mostly english) and %T uses the client language which would be more useful (my opinion). That's why you need after the Phrase Name the client index. So the second client is required using %T.

More informations are here about this: https://wiki.alliedmods.net/Translations_(SourceMod_Scripting)#Usage_in_a_Plugin

addons/sourcemod/scripting/surftimer/sql.sp
```diff
@@ -3458,13 +3441,388 @@ public void db_deleteCheckpoints()
...
+public void db_UpdatePRinfo_WithRuntime(int client, char szSteamID[32], int zGroup, float runtime)
+{	
+	if (!IsValidClient(client))
 		return;
-	}
+
+	char szQuery[2048];
+	//PrintToConsole(client, "%f || %f || %f || %f || %f\n", g_fTimeinZone[client][zGroup], g_fCompletes[client][zGroup], g_fAttempts[client][zGroup], g_fstComplete[client][zGroup], runtime);
+	Format(szQuery, 2048, sql_updatePrinfo_withruntime, g_fTimeinZone[client][zGroup], g_fCompletes[client], g_fAttempts[client][zGroup], g_fstComplete[client][zGroup], runtime, szSteamID, g_szMapName, zGroup);
+	SQL_TQuery(g_hDb, SQL_CheckCallback, szQuery, DBPrio_Low);
```
Bara, line 3814: you're using `DBPrio_Low` as data here instead of prio. You can use `.prio=DBPrio_Low` here.

Bara (changes requested):
Previous requests seems to be changed, I'm fine with `%t` for translations and reverted the change bacl. I also added a database upgrade for the new table.

Bara (approved):
Looks okay now When someone can confirm that everything works fine, then we can merge this. 👌🏻

shipyy:
i used `SQL_TQuery(g_hDb, SQL_CheckCallback, szQuery, DBPrio_Low);` w/o `.prio=DBPrio_Low` since i was going by code already written and i dont think i ever saw that on the code TIL :)

Bara:
> i used `SQL_TQuery(g_hDb, SQL_CheckCallback, szQuery, DBPrio_Low);` w/o `.prio=DBPrio_Low` since i was going by code already written and i dont think i ever saw that on the code TIL :)

exist code isn't always right 🙃 surftimer requires highly an update for some corrections like this. i already did a lot in the `oldDev` branch, but in some cases I made some mistakes and procudes another bugs from bug fixes.

PR #425, opened 2022-05-19
Native_GetPlayerInfo - add bonus level

tslashd:
Added a return for which bonus the client is currently on.
Small fixes for g_iInBonus[client]

### DIFF
--- a/addons/sourcemod/scripting/SurfTimer.sp
+++ b/addons/sourcemod/scripting/SurfTimer.sp
@@ -1939,6 +1939,7 @@ public void OnClientConnected(int client)
 	g_Stage[0][client] = 1;
 	g_bWrcpTimeractivated[client] = false;
 	g_CurrentStage[client] = 1;
+	g_iInBonus[client] = 0;
 }
 
 public void OnClientPutInServer(int client)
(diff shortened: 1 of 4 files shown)


addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -1185,7 +1185,8 @@ public Action Command_ToBonus(int client, int args)
 		zoneGrp = 1;
 
 	g_bInBonus[client] = true;
-	g_iInBonus[client] = zoneGrp;
+	if (g_mapZoneGroupCount > zoneGrp)
```
tslashd, line 1188: This stops a bug where if you type `/b <anyNumber>` that number will be assigned to `g_iInBonus[client]`

Have not noticed it breaking something on live version, but now this variable does not get appended with a new number if there are not that many bonus zones.

addons/sourcemod/scripting/SurfTimer.sp
```diff
@@ -1939,6 +1939,7 @@ public void OnClientConnected(int client)
 	g_Stage[0][client] = 1;
 	g_bWrcpTimeractivated[client] = false;
 	g_CurrentStage[client] = 1;
+	g_iInBonus[client] = 0;
```
tslashd, line 1942: Was throwing errors in server console until client reaches a StartZone iirc as this value was being set after they touch a start zone

tslashd: image



**N I C E**

addons/sourcemod/scripting/surftimer/commands.sp
tslashd: image



**N I C E**

Bara (approved):

PR #423, opened 2022-05-18
CSD fix 

shipyy:
Sry for closing #421 (im noob at github and fucked everything xD), but i have fixed things after. 
this request has the changes requested in #421 made to it, if anything more needs to be done just tell me :)

Close #349

PR #426, opened 2022-05-26
CSD customization

shipyy:
Since #423 was done, i tought some little customization would be cool aswell
Screenshot_4
l
* Separated CSD into its own menu
* Added Custom Color Option (RGB)
* Added Custom Position
* Added CSD Update Rate


shipyy:
Gonna PR with another branch instead (less conflicts)

PR #429, opened 2022-06-03
Update convars.sp

pedrotski:
Excessive FCVAR_NOTIFY in convars.sp is causing a timeout issue in SM 1.11.

[NET] Cannot send 5171-byte packet to 208.113.198.125:54479.  MTU is 1200

I have tested and indeed this fixes the issue. I don't think you would have this problem with SM 1.10 and A2S ruleshax as it patches the limit to 65000, SM 1.11 by default is 5000 now.

I would like to add, that I'm not a dev, things are probs wrong here but I made sure it compiles with no warnings or errors.


pedrotski:
I would like to add, that I'm not a dev, things are probs wrong here but I made sure it compiles with no warnings or errors.

pedrotski:
Whoops, wrong file, ill upload the correct one in another PR sorry.

PR #432, opened 2022-06-07
Update README

dPexxIAM:
readme was outdated, needed update.


Bara:
As I wrote in discord, the "Discord API" under credits/extensions can be removed too

dPexxIAM:
> As I wrote in discord, the "Discord API" under credits/extensions can be removed too

done, also added some of the FAQ to README. better suited here imo.

PR #433, opened 2022-06-08
Fix tables upgrades which stop after the first upgrade

Bara:
This is not tested.


freakexeuLow:
freak confirms it works :)

dPexxIAM (approved):

PR #438, opened 2022-06-13
Fix Player Titles

MSWS:
A [recent commit](https://github.com/surftimer/SurfTimer/commit/a7768e1ba4e64faa624cfe5a711ae8ceb4b682fb#r75944129) prevented players from setting a custom title if their custom title wasn't already enabled. This PR re-adds the insert statement if the database does not have any rows matching a Steam ID.


DaLLuZZ (approved):
That will work

Bara (approved):

PR #444, opened 2022-06-19
Replay cps

shipyy:
image

Allow players to choose from which checkpoint/stage they want to view a replay


dPexxIAM:
seems fine, nice addition. although i would prefer ck_checkpoints schema be remade to not use 60+ columns per row.

shipyy:
ck_checkpoints? or u mean ck_replays? :)

dPexxIAM:
> ck_checkpoints? or u mean ck_replays? :)

both :)

1zc:
I do feel like the current layout of `ck_replays` will be an issue down the line, especially with people wanting to remake `ck_checkpoints` to not have a limit, so it's an inevitable change to come if we don't get it out of the way now.

Might be better to instead introduce a column that tracks what checkpoint is being stored, could be part of the index too. Unsure if anyone else has any better ideas though, so maybe wait for more people to chime in 😄

tslashd:
imho reworking `ck_checkpoints` would make it a bit easier to work with the code (not have these long queries to set data in 35+ columns), should also make running `classics3` easier to set up, just by editing max cp limit in `SurfTimer.sp` 
also maybe a bit better performance as it will not get that much column data from the table when sending a query

dPexxIAM (approved):

addons/sourcemod/scripting/surftimer/sql.sp
```diff
@@ -3412,6 +3424,97 @@ public void SQL_selectCheckpointsCallback(Handle owner, Handle hndl, const char[
...
 
+public void db_viewReplayCPTicks(char szMapName[128])
+{
+	char szQuery[1024];
+
+	for(int style = 0; style < MAX_STYLES; style++){
+		g_bReplayTickFound[style] = false;
+
+		Format(szQuery, sizeof(szQuery), sql_selectReplayCPTicksAll, szMapName, style);
+		SQL_TQuery(g_hDb, SQL_selectReplayCPTicksCallback, szQuery, DBPrio_Low);
```
Bara, line 3435: so btw. `DBPrio_Low` is here `data` which shouldn't be the logic here. https://sourcemod.dev/#/dbi/function.SQL_TQuery

PR #445, opened 2022-06-19
Stage sm_startpos

shipyy:
Allows players to use the cmd ```sm_startpos``` in stage zones


DaLLuZZ:
This will lead to lots of cheating srcp times on many stages on so many maps...

tslashd:
Yeah using this would make stage records more abusable than they currently are (staged linears like ace)

Would have to use maybe prac timer when doing a stage after using `sm_startpos` imo

shipyy:
@DaLLuZZ @tslashd i made so u can only use startpos for stages when not in a run

DaLLuZZ:
Do you have a test server?

shipyy:
locally xd

tslashd:
> @DaLLuZZ @tslashd i made so u can only use startpos for stages when not in a run

So when you're in a map run will you get teleported to the default stage spawn or wherever you've set your `startpos` for the given stage?

shipyy:
if u are in a run u will teleported to the default stage position

shipyy:
im asuming u mean like using "sm_rs" during a run? because ye i did take that into account :D

dPexxIAM (approved):
ty

PR #446, opened 2022-06-19
saveloc ranks 

shipyy:
image

Displays the rank of the time in map/bonus/stages

### DIFF
--- a/addons/sourcemod/scripting/SurfTimer.sp
+++ b/addons/sourcemod/scripting/SurfTimer.sp
@@ -852,6 +852,9 @@ float g_fCurrentRunTime[MAXPLAYERS + 1];
 // PracticeMode total time the run took in 00:00:00 format
 char g_szPracticeTime[MAXPLAYERS + 1][32];
 
+//Rank of the PracticeMode time
+int g_iPracRunTimeRank[MAXPLAYERS + 1];
+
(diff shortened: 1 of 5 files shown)


dPexxIAM:
I believe we can use existing functions here instead of making new ones entirely for prac mode? We have SQL_CurrentRunRankCallback, db_viewBonusRunRank, g_StageRank

shipyy:
i decided to separate each, because the functions/variables u are mentioning are used when not in practice mode, so i just split them up since i tought would be more organized and clean instead of mixing all the prac/non prac stuff in the same functions

dPexxIAM:
> i decided to separate each, because the functions/variables u are mentioning are used when not in practice mode, so i just split them up since i tought would be more organized and clean instead of mixing all the prac/non prac stuff in the same functions

We have the global to check whether the player is in practice mode or not, why do we need to make new functions?

shipyy:
@dPexxIAM did wat u requested :), except for the stage prac which it is required to get the run stage rank and not the players best stages rank which i believe is what g_stagerank does? xD

dPexxIAM:
> @dPexxIAM did wat u requested :), except for the stage prac which it is required to get the run stage rank and not the players best stages rank which i believe is what g_stagerank does? xD

👍

PR #447, opened 2022-06-19
CP speed comparisons

shipyy:
image

Displays speed comparisons in map checkpoints/stages


dPexxIAM:
before this is merged i would prefer if ck_checkpoints schema is reworked, as discussed on the discord previously

Bara:
I've marked this as draft, until checkpoints table conversion was merged.

shipyy:
i believe since we did ck_checkpoints restructure this type of functionality could be implemented differently and in better ways i suggest closing this :)

PR #451, opened 2022-06-21
remove simple stripper fixes

dPexxIAM:
probably won't wanna hard code stripper fixes into the timer lol


Bara (approved):
definitely not needed when we can do it with stripper.

Issue #452, opened 2022-06-23
[BUG] Bonus Style Menu only showing Normal Style


MSWS:
**Describe the bug**
The record menu currently does not support viewing top records of a specific style in a certain bonus. 
The records of the Normal Style is shown for the selected bonus stage.

**Replication**
(Be on a map with at least 1 bonus stage)
1. Type !top
2. Pick a style that is not normal
3. Press Bonus Top
4. (If applicable) Select the bonus

**Expected behavior**
The records of the specified style should be shown for the selected bonus stage

PR #453, opened 2022-06-23
Support bonus styles (#452)

MSWS:
This merge request fixes:

The record menu currently does not support viewing top records of a specific style in a certain bonus. 

Replication:
Type !top
Pick a style that is not normal
Press Bonus Top
(If applicable) Select the bonus

Bug: The records of the Normal Style is shown for the selected bonus stage
Desired: The records of the specified style should be shown for the selected bonus stage


MSWS:
https://github.com/surftimer/SurfTimer/issues/452

dPexxIAM (approved):

Issue #465, opened 2022-07-05
[BUG] Exiting zone not start timer


daffyyyy:
**Describe the bug**
Exiting the zone will not start the timer, when the zone will shrink then yes. Making a smaller zone on the edge doesn't work either

**Expected behavior**
The timer should start after leaving the zone

**Screenshots**
https://i.imgur.com/hAKusF6.mp4

**Plugin List**
```  01 "PushFix - Definitive Edition" (1.0.0) by Original idea xutaxkamay | Implementation GAMMACASE
  02 "SurfTimer Vote Extend" (1.8.3) by SurfTimer Contributors
  03 "HeadBugFix" (1.0.0) by GAMMA CASE
  04 "Basic Ban Commands" (1.10.0.6544) by AlliedModders LLC
  05 "Basic Votes" (1.10.0.6544) by AlliedModders LLC
  06 "Fun Votes" (1.10.0.6544) by AlliedModders LLC
  07 "EndTouch Fix" (1.0) by rumour, mev
  08 "SurfTimer" (1.0.dev) by All contributors
  09 "Auto exec server.cfg" (1.1) by MR. EMINEM
  10 "Basic Commands" (1.10.0.6544) by AlliedModders LLC
  11 "Player Commands" (1.10.0.6544) by AlliedModders LLC
  12 "ServerAdvertisements3" (3.3.0) by ESK0
  13 "Admin Help" (1.10.0.6544) by AlliedModders LLC
  14 "Weapon Econ Team Transfer" (1.1) by Wend4r
  15 "SurfTimer MapChooser" (2.0.0) by AlliedModders LLC & SurfTimer Contributors
  16 "Momentum surf fix '2" (1.1.3) by GAMMA CASE
  17 "Reserved Slots" (1.10.0.6544) by AlliedModders LLC
  18 "Admin Menu" (1.10.0.6544) by AlliedModders LLC
  19 "Anti-Flood" (1.10.0.6544) by AlliedModders LLC
  20 "[Surf Timer] Teleport Destination Finder" (1.0) by marcowmadeira
  21 "CS:GO Movement Unlocker" (1.0) by Peace-Maker
  22 "SurfTimer Rock The Vote" (2.0.0) by AlliedModders LLC & SurfTimer Contributors
  23 "Normalized Run Speed" (0.6) by sneaK
  24 "Client Preferences" (1.10.0.6544) by AlliedModders LLC
[...]

PR #466, opened 2022-07-05
remove utopia_njv zones

dPexxIAM:


dPexxIAM:
https://github.com/surftimer/SurfTimer/issues/465 will come up again unless zones are removed

Bara (approved):

PR #482, opened 2022-07-25
PR info bug fix

shipyy:
The variable  `runtime` was being set with wrong values


dPexxIAM (changes requested):
no client index? :D

shipyy:
@dPexxIAM this float `fCurrentRunTime = g_fCurrentRunTime[client];` is set beforehand :)

dPexxIAM:
> @dPexxIAM this float `fCurrentRunTime = g_fCurrentRunTime[client];` is set beforehand :)

right, which is pointless and confusing. don't you agree?

Bara (approved):

dPexxIAM (approved):

PR #483, opened 2022-07-26
dont reload ck_chat_record_type

tslashd:
Currently `ck_chat_record_type` goes back to `0` which is default value set in code if the map reloads `mp_restartgame`


tslashd:
Also I do not see the reasoning behind applying that value on every round start as it's a value we have already imported from `surftimer.cfg` by the time we get to `OnMapStart` or `Event_OnRoundStart`

Bara (approved):

PR #484, opened 2022-07-29
Country top

shipyy:
# Allow player's to see the leaderboards of any country in every style

## sm_ctop
image
image

## sm_crank
image

allows multiple inputs aswell i.e :
` sm_crank <playername> `
` sm_crank <style> `
` sm_crank <playername> <style> `
` sm_crank <style> <playername> `


tslashd:
For some reason it's saying the players are from Bulgaria in chat, but they are not:
crank country


`sm_crank <style>` - seems to not work for all styles and it works only if you type the full string of the style (got it to print something for `Freestyle, Low-Gravity, Normal` nothing else)

`<style>` var if possible should be able to use abbreviations (`Freestyle = FS; Low-Gravity = LG; Half-Sideways = HSW;` etc.) for the styles as currently you need to write the whole style name like I have shown above.
Almost nobody knows the `int` corresponding to the style in order to use that

addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -1781,6 +1787,358 @@ public Action Command_ckNoClip(int client, int args)
...
+
+			char arg1[MAX_NAME_LENGTH];
+
+			GetCmdArg(1, arg1, sizeof arg1);
+
+			//sm_ctop #<style_acronym>
+			if (arg1[0] == '#') {
+				ReplaceString(arg1, sizeof arg1, "#", "", false);
+
+				ArrayList styles = new ArrayList(32);
```
Bara, line 1844: ArrayList must be deleted, otherwise you'll run sometime into an memory leak.

addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -1781,6 +1787,358 @@ public Action Command_ckNoClip(int client, int args)
...
+		szCountryName = "none";
+	}
+
+
+	Menu menu = CreateMenu(CountryTopMenuStyleSelectHandler);
+
+	SetMenuTitle(menu, "Top Menu - Select a style\n \n");
+
+	//EACH MENU ITEM HAS A STRING CONTAINING THE COUNTRY NAME AND THE STYLE SELECTED IN THE FOLLOWING FORMAT -> 'country-style'
+	Format(szItem, sizeof(szItem), "Normal");
```
Bara, line 2084: We could loop `g_EditStyles` here, or? Would save us some lines of code.

Bara (changes requested):

Bara (approved):
Would be nice if someone could test this ( @tslashd maybe? )

tslashd:
> Would be nice if someone could test this ( @tslashd maybe? )

Missing `}` in [translation](https://github.com/shipyy/Surftimer-Official/blob/b38fe9a60aacff5c6aa2d8a4f7a3fb3ade5361ea/addons/sourcemod/translations/surftimer.phrases.txt#L1872) file was throwing error.
Other than than all seems good

Issue #486, opened 2022-08-05
[Suggestion] !nextrank command


MSWS:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Using the !nextrank would tell you how many points you would need to reach the next rank. Players are always asking what rank (or how many points) they need to reach the next rank. The command would check their currents points (or rank) and display this and check how many points they would need to reach the next rank (even if their next rank is based on rank and not points it should Always display in points as this is the format that players are most familiar with). It would look something like "Next rank up in X points.".

**Additional context**
Originally suggested internally.

PR #487, opened 2022-08-05
sm_nextrank

shipyy:
#486 
displays the amount of points required to achieve next rank
image

also prints msg if the player is already max rank
image


MSWS:
Woah, that's cool! Thanks for adding this!

Bara (approved):

PR #491, opened 2022-08-07
Add country and continent codes to ck_playerrank

Bara:
**Important Note**: _With this change we enforce the usage of SM1.11 and higher._


Bara:
Compile error will be fixed with #461

dPexxIAM (approved):

shipyy:
@Bara could add full continent name aswell, like with the countries.
would make formatting a bit easier :)

PR #492, opened 2022-08-07
wrong index v2

shipyy:
wrong index was because i forgot that bonuses are inserted to ck_checkpoints


tslashd:
think I tested this last night with @shipyy 🤔

dPexxIAM (approved):

PR #496, opened 2022-08-09
wait 1 tick before disabling practice mode

dPexxIAM:
without this clients can possibly set a "legit" time by pressing !r when they reach endzone

### DIFF
--- a/addons/sourcemod/scripting/surftimer/misc.sp
+++ b/addons/sourcemod/scripting/surftimer/misc.sp
@@ -4708,6 +4708,8 @@ public Action DisablePrac(Handle timer, any data)//saveloc on start > startpos
 	int client = GetClientFromSerial(data);
 	
 	g_bPracticeMode[client] = false;
+
+	return Plugin_Handled;
 }
 
 public void ReadDefaultTitlesWhitelist()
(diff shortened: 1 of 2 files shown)


addons/sourcemod/scripting/surftimer/misc.sp
```diff
@@ -4703,6 +4703,15 @@ public void TeleportToSaveloc(int client, int id)
 	TeleportEntity(client, g_fSaveLocCoords[client][id], g_fSaveLocAngle[client][id], g_fSaveLocVel[client][id]);
 }
 
+public Action DisablePrac(Handle timer, any data)//saveloc on start > startpos
+{
+	int client = GetClientFromSerial(data);
```
Bara, line 4708: Make sure that the client is still valid

Bara (changes requested):

addons/sourcemod/scripting/surftimer/misc.sp
dPexxIAM: > Make sure that the client is still valid

this only gets run through client commands. there exist valid client checks there.

Bara (approved):

PR #498, opened 2022-08-09
Fix Replays with CP selected

shipyy:
error occuring before 
image

was due to wrong query format resulting in a false positive regarding duplicate entry

added a check to not display the starting points from replay incase there was no replay cp ticks found when loading

also swapped multiple queries to a unique transaction instead

### DIFF
--- a/addons/sourcemod/scripting/surftimer/sql.sp
+++ b/addons/sourcemod/scripting/surftimer/sql.sp
@@ -3443,7 +3443,8 @@ public void db_UpdateReplaysTick(int client, int style){
 		}
 	}
 
-	SQL_ExecuteTransaction(g_hDb, TicksTransaction, db_TicksTransactionOnSuccess, db_TicksTransactionOnFailure, DBPrio_Low);
+	SQL_ExecuteTransaction(g_hDb, TicksTransaction, db_TicksTransactionOnSuccess, db_TicksTransactionOnFailure, _, DBPrio_Low);
+
 }
 
 public void db_TicksTransactionOnSuccess(Handle db, any pack, int numQueries, Handle[] results, any[] queryData)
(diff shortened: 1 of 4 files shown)


tslashd:
Error seems to not be present after this.

_Before this fix the replay from the map mentioned in PR has been replaced by a non WR tho_

addons/sourcemod/scripting/surftimer/sql.sp
```diff
@@ -3427,28 +3427,33 @@ public void db_UpdateReplaysTick(int client, int style){
...
 	}
 	else{
 		for(int i = 0; i < cp_count; i++){
 			Format(szQuery, sizeof(szQuery), sql_updateReplayCPTicks, g_szMapName, i+1, g_iCPStartFrame[style][i], style);
-			SQL_TQuery(g_hDb, SQL_UpdateReplaysTickCallback, szQuery, _, DBPrio_Low);
+			TicksTransaction.AddQuery(szQuery);
 		}
 	}
+
+	SQL_ExecuteTransaction(g_hDb, TicksTransaction, db_TicksTransactionOnSuccess, db_TicksTransactionOnFailure, DBPrio_Low);
```
Bara, line 3446: DBPrio_Low is here wrong. Replace it with this:
`SQL_ExecuteTransaction(g_hDb, TicksTransaction, db_TicksTransactionOnSuccess, db_TicksTransactionOnFailure, _, DBPrio_Low);`

Bara (changes requested):

addons/sourcemod/scripting/surftimer/sql.sp
Bara: or `.priority=DBPrio_Low` works too

tslashd:
does not update CPs in chat until reconnecting to the server

shipyy:
the errors related to ck_checkpoints were happening because the changes were not merged from the other PR(#499 ) that fixes taht issue

Bara (approved):

PR #499, opened 2022-08-09
db_UpdateCheckpoints() fix

shipyy:
error prior to fix
image

issues were:
* query to update had wrong column name `cp` instead of `time`
* query size was too low
* query formating had missing 'cp' value

thx @tslashd for helping out on this (:


tslashd:
No errors are being thrown around upon improving your PB or setting one with this

Bara (approved):

PR #505, opened 2022-08-21
more api

shipyy:


addons/sourcemod/scripting/include/surftimer.inc
```diff
@@ -101,11 +101,12 @@ native void surftimer_GetPlayerNameColored(int client, char[] str);
...
  *
- * @param name				String to store the WR holder's name.
- * @param time				String to store the WR holder's time.
+ * @param szname				String to store the WR holder's name.
+ * @param sztime				String to store the WR holder's time.
+ * @param time					float to store the WR holder's time.
  * @return						The number of players who have finished the current map.
  */
-native int surftimer_GetMapData(char[] name, char[] time);
+native int surftimer_GetMapData(char[] szname, char[] sztime, float& time);
```
Bara, line 109: What's the reason for this change? Plugins which use the old version, will not work anymore.

addons/sourcemod/scripting/include/surftimer.inc
```diff
@@ -117,7 +118,7 @@ native int surftimer_GetMapData(char[] name, char[] time);
  * @param countryCode			String to store the client's countryCode.
  * @param continentCode			String to store the client's continentCode.
  */
-native void surftimer_GetPlayerData(int& client, char[] time, int& rank, char[] country, char countryCode[3], char continentCode[3]);
+native void surftimer_GetPlayerData(int& client, float& time, int& rank, char[] country, char countryCode[3], char continentCode[3]);
```
Bara, line 121: What's the reason for this change? Plugins which use the old version, will not work anymore.

addons/sourcemod/scripting/include/surftimer.inc
```diff
@@ -151,15 +168,15 @@ native bool surftimer_SafeTeleport(int client, float fDestination[3], float fAng
  * @param client			The client's ID.
  * @return						Plugin_Handled, to stop the client from searching.
  */
-forward Action surftimer_OnMapFinished(int client, float fRunTime, char sRunTime[54], int rank, int total, int style);
+forward Action surftimer_OnMapFinished(int client, float fRunTime, char sRunTime[54], float PBDiff, float WRDiff, int rank, int total, int style);
```
Bara, line 171: What's the reason for this change? Plugins which use the old version, will not work anymore.

addons/sourcemod/scripting/include/surftimer.inc
```diff
@@ -151,15 +168,15 @@ native bool surftimer_SafeTeleport(int client, float fDestination[3], float fAng
...
 
 /**
  * Called when a player reaches a checkpoint zone or a stage.
  *
  * @param client			The client's ID.
  * @return						Plugin_Handled, to stop the client from searching.
  */
-forward Action surftimer_OnCheckpoint(int client, float fRunTime, char sRunTime[54], float fPbCp, char sPbDiff[16], char fSrCp, char sSrDiff[16]);
+forward Action surftimer_OnCheckpoint(int client, float fRunTime, char sRunTime[54], float fPbCp, char sPbDiff[16], float fSrCp, char sSrDiff[16]);
```
Bara, line 179: What's the reason for this change? Plugins which use the old version, will not work anymore.

addons/sourcemod/scripting/include/surftimer.inc
```diff
@@ -168,7 +185,7 @@ forward Action surftimer_OnCheckpoint(int client, float fRunTime, char sRunTime[
  * @param client			The client's ID.
  * @return						Plugin_Handled, to stop the player from searching.
  */
-forward Action surftimer_OnBonusFinished(int client, float fRunTime, char sRunTime[54], int rank, int total, int bonusid);
+forward Action surftimer_OnBonusFinished(int client, float fRunTime, char sRunTime[54], float fPBDiff, float fSRDiff, int rank, int total, int bonusid, int style);
```
Bara, line 188: What's the reason for this change? Plugins which use the old version, will not work anymore.

Bara (commented):
I just saw this changes and don't really want to look at the rest of the changes without knowing the reasons for the native/forward changes due incompatibility changes with exist plugins.

addons/sourcemod/scripting/include/surftimer.inc
shipyy: having the value itself seems to be a better practice, it can be manipulated much easier instead of begin restricted to a formatted string

addons/sourcemod/scripting/include/surftimer.inc
shipyy: having the value itself seems to be a better practice, it can be manipulated much easier instead of begin restricted to a formatted string

addons/sourcemod/scripting/include/surftimer.inc
shipyy: this one i beleive it was wrong, the forward  has `call_pushfloat()`

addons/sourcemod/scripting/include/surftimer.inc
shipyy: this was intented to proved the time differences of the player.
i added it to this forward since it made sense for me to have the variables in the forward that gets called when the player finishes a run

addons/sourcemod/scripting/include/surftimer.inc
shipyy: same reason as `onmapfinished`

Bara (approved):

PR #509, opened 2022-08-27
Improved backwards style checks

Synert:
Updated backwards checks to only reset style if keys pressed and going above 260 units. Prevents style resets when hitting walls or climbing ladders


Sayt123:
Tested on Surf Central Station and haven't run into any issues with the change.

addons/sourcemod/scripting/surftimer/hooks.sp
```diff
@@ -940,28 +943,36 @@ public Action OnPlayerRunCmd(int client, int &buttons, int &impulse, float vel[3
...
+			if (g_KeyCount[client] < 59)
+				g_KeyCount[client]++;
+			else
 			{
-				g_iCurrentStyle[client] = 0;
-				g_KeyCount[client] = 0;
-				CPrintToChat(client, "%t", "Hooks14", g_szChatPrefix);
+				// don't check if we're below the threshold, or not pressing anything.
+				// this should solve resets happening on wall collisions and ladders
+				if(bInputs && fSpeedSqr > fSpeedThres * fSpeedThres)
```
Bara, line 966: `fSpeedThres * fSpeedThres` is this correct?

Synert: Yeah, since the speed is squared and the threshold isn't

dPexxIAM (approved):

PR #512, opened 2022-08-28
attempt at multiple zones :)

shipyy:
i tested with quite some different maps and seemed to be working as intended, but ya never know


shipyy:
uh oh  i did a oopsie lmao

Issue #524, opened 2022-09-12
[BUG] Style times not being recorded if one already exists


Synert:
If no time has been set, then everything works as expected. If a time is already set, everything else works as expected (replay bot updates, messages are printed, points given) but the time itself will not be saved. Observed while using 1.1.2-beta1 on servers SCEN and Thunderdome.

The bug appears to be caused by #501 and is fixed by undoing this change. This does cause [this Discord bug](https://github.com/surftimer/SurfTimer-discord/issues/21) to reappear, though.

Bara:
Fixed in dev branch

PR #525, opened 2022-09-12
fix for #524

shipyy:
previous pr fix for this in hindsifghted i believe was missing a few more things (til)

* fix for wrong PERSONAL record query
* fix for wrong position to send new record forward

### DIFF
--- a/addons/sourcemod/scripting/surftimer/buttonpress.sp
+++ b/addons/sourcemod/scripting/surftimer/buttonpress.sp
@@ -487,6 +487,7 @@ public void CL_OnEndTimerPress(int client)
 				{
 					// New fastest time in map
 					g_bStyleMapSRVRecord[style][client] = true;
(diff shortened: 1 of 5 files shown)


shipyy:
i have not tested it (cannot atm) if someone else could confirm it does in fact work would be awesome! (:

Synert:
Style times are saving properly, but records are not being output in the Discord or printing the record message (<player> beat the <style> map record). This is also occurring for the first time set

Synert (changes requested):
Typo in surftimer.phrases.txt - '{4,s}' -> '{4:s}'. Other changes confirmed working on Thunderdome

shipyy:
hopefully all good now (:

Synert (approved):
Tested and working as expected on Thunderdome

dPexxIAM (approved):

PR #538, opened 2022-11-14
add mapper names to ck_maptier

dPexxIAM:
adds mapper name column to ck_maptier


addons/sourcemod/scripting/surftimer/admin.sp
```diff
@@ -51,6 +51,33 @@ public int TopMenuHandler2(Handle topmenu, TopMenuAction action, TopMenuObject o
...
+
+	if (!IsPlayerZoner(client))
+	{
+		CReplyToCommand(client, "%t", "AdminMapperName", g_szChatPrefix);
+		return Plugin_Handled;
+	}
+
+	if (args == 0)
+	{
+		CReplyToCommand(client, "\x05GF \x01| Usage: sm_addmappername <String>");
```
Bara, line 67: Prefix?

addons/sourcemod/scripting/surftimer/globals.sp
```diff
@@ -1152,6 +1152,12 @@ char g_szCountry[MAXPLAYERS + 1][100];
 char g_szCountryCode[MAXPLAYERS + 1][3];
 char g_szContinentCode[MAXPLAYERS + 1][3];
 
+// mappers name
+char g_szMapperName[32];
+
+// mapper name null?
+bool g_bMapperNameFound;
```
Bara, line 1159: You could check the string length instead of this bool

dPexxIAM: maybe, but i don't see why we can't just use the bool

Bara (approved):

PR #544, opened 2022-12-11
ck_prestige_rank fix

dPexxIAM:
fix for sql_selectRankedPlayersRankCallback having no result (no rank) players would be kicked from the server if the convar was set to 0


tslashd:
Just tested. 
Unranked players are no longer kicked when having this set to **0**, thanks for fixing <3

Without this PR:
image

Bara (approved):

PR #547, opened 2022-12-27
Missing mapper name argument for Map Info

tslashd:
**/mi** was erroring out cause of missing mapper name argument


dPexxIAM (changes requested):
You don't need to change the query and define the new string, we already have the global g_szMapperName filled, just use that.

dPexxIAM (approved):

PR #554, opened 2023-02-11
Time formatting

tslashd:
Nothing much really, I think it makes times easier to read and also looks better


sneak-it (dismissed):
Using a `.` instead of `:` is more accurate, +1

lgtm

tslashd:
forgot about the hardcoded `00:00:00` when timer is stopped

1zc (approved):
Good stuff, lgtm!

PR #558, opened 2023-03-05
CPR command update

tslashd:
With this PR players can now use `/cpr @<rank> @<rank> <mapname>` as well as `/cpr @<rank> <in-game Player name> <mapname>` where mapname is optional on all.

**Currently comparing 2 players with their names works only for the current map.**
Also comparing by name has not been altered from original (only looks for in-game players), only changed part is to include client that executed the command in the search otherwise client could not compare themselves with another player/rank by using their name.

~~Currently compiling this will throw warnings for unused variables:~~
image
~~The "fix" is to uncomment the `SetPackPosition` and `DataPackPos` functions and remove the unused variables, the reason I have not done that is cause we would have to hardcode positions for the DataPack.~~
Fixed [here](https://github.com/surftimer/SurfTimer/pull/558/commits/eabefef34d5d31a9c2e192eed63c6a3ede04b07d) and [here](https://github.com/surftimer/SurfTimer/pull/558/commits/41084dfbebbea073c6ffdc5ac01a26332eb7d058)

Current known issues:
1. ~~Running the command on a map like `surf_beginner` will show `surf_beginner_lin` (in my case) cause of `LIKE` in the query~~ Fixed [here](https://github.com/surftimer/SurfTimer/pull/558/commits/2415613162349a6464eb32f623922a0d09eab5e6)

### DIFF
--- a/addons/sourcemod/translations/surftimer.phrases.txt
+++ b/addons/sourcemod/translations/surftimer.phrases.txt
@@ -856,6 +856,11 @@
 		"#format"	"{1:s}"
 		"en"		"{1} New stage start position saved"
 	}
+	"Commands90"
+	{
+		"#format"	"{1:s}"
+		"en"		"{1} You {darkred}do not{default} have a personal best yet, use:\n{1} !cpr {green}@<rank> @<rank> {blue}<mapname> {default}(mapname is optional)"
+	}
 	"BotInUse"
 	{
 		"#format"	"{1:s},{2:s}"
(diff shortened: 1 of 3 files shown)


addons/sourcemod/translations/surftimer.phrases.txt
```diff
@@ -1982,4 +1987,5 @@
 		"#format"	"{1:s}"
 		"en"		"{1} {yellow}Already Max Rank!"
 	}
+	""
```
Bara: Huch?

dPexxIAM (approved):
looks good

PR #562, opened 2023-03-19
[ADDITION] OnCheckpoint forward

tslashd:
Add sending the `Checkpoint` number to the forward.
Marking as draft for now in order to add some more information to the forward in the `include` file and maybe test it? 👀


tslashd:
Will affect [Clean-HUD](https://github.com/shipyy/clean-hud/blob/main/scripting/chud-forwards.sp#LL1) plugin, will PR change there as well when this one is merged.

If any other plugins would be affected, please leave a comment or PR the change, thanks.

dPexxIAM (approved):

PR #566, opened 2023-03-21
Fix /hide and a few other player SDKHooks

1zc:
Currently facing a weird issue with the dev branch - seems like /hide no longer works as expected. I did some digging and it seems like calling SDKHooks after the SteamID validation stuff in `OnPlayerPutInServer` is causing the issue. 

Not entirely sure exactly why that happens, but I moved the SDKHook calls above the SteamID validation code in the function. I've also added a new bool to try and ensure that the SDKHook functions get called only once, as a safeguard. (Not sure if this is truly required, but if it isn't, do let me know and I'll remove it)

Tested working on GFL and on the NA SurfTimer test server

### DIFF
--- a/addons/sourcemod/scripting/SurfTimer.sp
+++ b/addons/sourcemod/scripting/SurfTimer.sp
@@ -416,6 +416,16 @@ public void OnClientPutInServer(int client)
 		return;
 	}
 
+	// SDKHooks
+	if (g_bClientHooksCalled[client] == false)
+	{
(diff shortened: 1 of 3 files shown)


tslashd:
I believe this PR also fixes a recent issue where replay bots were taking fall damage.

Tested and `!hide` works fine, no replay bots were harmed in the process of testing 😆

Bara (approved):

Issue #565, opened 2023-03-21
[BUG] Bots Repeatedly Dying, Memory Leak, Crashed Server


IG-PhantomX:
**Describe the bug**
Not exactly sure what the problem was with these particular replays - on map load to surf_hearth_fix, bots repeatedly died.  The server error log spammed the following about 10000 times, then crashed the server.

L 03/20/2023 - 21:27:52: [SM] Exception reported: Invalid index 0 (count: 0)
L 03/20/2023 - 21:27:52: [SM] Blaming: SurfTimer.smx
L 03/20/2023 - 21:27:52: [SM] Call stack trace:
L 03/20/2023 - 21:27:52: [SM]   [0] GetArrayCell
L 03/20/2023 - 21:27:52: [SM]   [1] Line 994, surftimer\surfzones.sp::BeamBox_OnPlayerRunCmd
L 03/20/2023 - 21:27:52: [SM]   [2] Line 1110, surftimer\hooks.sp::OnPlayerRunCmd

Perhaps there should be a mechanism in place that if there is a faulty replay, the bots just stay dead and sends 1 error report.

IG-PhantomX:
The memory leak is solved, but still have an issue with the bots dying.  This is after implementing all the most recent code modifications.  Will start new post with video on this.

PR #567, opened 2023-03-21
[FIX] BeamBox_OnPlayerRunCmd

tslashd:
**_Not thoroughly tested._** (new runs are recorded, saved and replayed as expected)
Not really sure why we are sending zones drawing data to bots.

Fixes the error from #565.

### DIFF
--- a/addons/sourcemod/scripting/surftimer/hooks.sp
+++ b/addons/sourcemod/scripting/surftimer/hooks.sp
@@ -1106,8 +1106,8 @@ public Action OnPlayerRunCmd(int client, int &buttons, int &impulse, float vel[3
 
 		g_fLastSpeed[client] = speed;
 		g_LastButton[client] = buttons;
-
-		BeamBox_OnPlayerRunCmd(client);
+		if (!IsFakeClient(client))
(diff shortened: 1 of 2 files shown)


sneak-it (changes requested):
So it seems like the only instance of `BeamBox_OnPlayerRunCmd` is called is here:
https://github.com/surftimer/SurfTimer/blob/76955ababa8e0f5a2b6bae83e0b26dbf2d963627/addons/sourcemod/scripting/surftimer/hooks.sp#L1110

`OnPlayerRunCmd` is only called when client movement buttons are being pressed - console does not have movement buttons, so checking if `client == 0` is unnecessary. In the case of a listen server being used (which is unsupported by SM in general), if this plugin magically works, it would break this feature as well.

In regards to the other bot checks, these are better encompassed by utilizing `IsFakeClient` check. This would cover any additional bots and futureproof renaming/removing/additions of existing bot globals. I would also encourage changing this check location to inside `OnPlayerRunCmd`, as it will no longer be required to enter this function entirely.

Good find!

sneak-it (approved):
LGTM! 👍

PR #577, opened 2023-04-08
[FIX] Some things

tslashd:
Fix the lag machine created by replay bots and their SteamIDs.
Fix restoring position from `ck_playertemp`


Bara:
I'm not pretty sure about this pull request. Should be tested before merging.

tslashd:
Resolved the conflicts.

I would recommend switching off `position restoring` as there are lots of edge cases that are not handled with that system.


Example:
1. What do we do if a `record` is set from a restored run?
    -  Where do we get replay data from, or checkpoints data, etc.

Bara (approved):

PR #581, opened 2023-04-15
Fixed Normal Mode Issues - CCarrMcMahon

seethingword:


addons/sourcemod/scripting/surftimer/commands.sp
```diff
@@ -623,18 +623,25 @@ public void VoteExtend(int client)
...
-	if (g_bPracticeMode[client])
-		g_bPracticeMode[client] = false;
+    if (g_bPracticeMode[client])
+        g_bPracticeMode[client] = false;
 
-	Command_Restart(client, 1);
+    g_iCurrentStyle[client] = 0;
+    g_iInitalStyle[client] = 0;
+    Format(g_szInitalStyle[client], 128, "Normal");
+    Format(g_szStyleHud[client], 32, "");
```
Bara, line 637: Why hardcoded values instead of using `sizeof(g_szInitalStyle[])` and `sizeof(g_szStyleHud[])`?

seethingword: It was a quick fix until we have more time to fix it. if those would work, go for it

CCarrMcMahon: > Why hardcoded values instead of using `sizeof(g_szInitalStyle[])` and `sizeof(g_szStyleHud[])`?

Because this was hardcoded in the original code. All I did was copy it over. Go check out the StyleSelectMenuHandler section in the commands.sp file.

https://github.com/surftimer/SurfTimer/blob/bd094ef0c78d139de3ce9fc9bd244e27c82b1c29/addons/sourcemod/scripting/surftimer/commands.sp#L4980-L5075

dPexxIAM (approved):

PR #583, opened 2023-04-23
On client put in server fix

dPexxIAM:
If steamID was not initialized we would return from this function and therefore necessary bools and settings would not be set leading to some issues


IG-PhantomX:
Tested this live on public servers.  This fixed issues related to false time being set (0.0 times), and also issues with godmode.

Bara (approved):

PR #591, opened 2023-07-12
[Addition] Add Checkpoints ArrayList for map finishes and map records

tslashd:
The **ArrayList** contains data _(Checkpoint Number, Style, PB Difference, WR Difference)_ for each checkpoint of the player run.

- Array is reset each time player touches a start zone

:BearPepe: 

### Will most likely break existing plugins using these forwards so please update your plugins!

### DIFF
--- a/addons/sourcemod/scripting/SurfTimer.sp
+++ b/addons/sourcemod/scripting/SurfTimer.sp
@@ -550,6 +550,7 @@ public void OnClientAuthorized(int client)
 public void OnClientDisconnect(int client)
 {
 	db_savePlayTime(client);
+	delete g_aCheckpointsDifference[client];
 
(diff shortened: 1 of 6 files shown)


addons/sourcemod/scripting/SurfTimer.sp
```diff
@@ -612,6 +615,7 @@ public void OnClientDisconnect(int client)
 				g_fTimeinZone[client][zonegroup] += g_fTimeIncrement[client][zonegroup];
 			db_UpdatePRinfo(client, g_szSteamID[client], zonegroup);
 		}
+		delete g_aCheckpointsDifference[client];
```
Bara: I think this line can be moved earlier inside this function. As example after `db_savePlayTime(client);`.

Bara (approved):

Issue #592, opened 2023-07-12
[Suggestion]


tmck-code:
First off, thanks so much for making this! This repo _is_ surf, allowed me to run my own private server super-easily and freed me from the shackles of public servers :heart: 

**Is your feature request related to a problem? Please describe.**

1. The prestrafe speed is displayed in the chat
    - The chat box is in the corner of the screen, and I find that this negatively impacts any prestrafe practice as my eyes have to dart back and forth
    - From what I've searched, the chat box cannot be moved
2. Momentum mod displays prestrafe speed underneath the central speed display
    I'm convinced that Momentum Mod will be in beta for a while yet

**Describe the solution you'd like**

Display the prestrafe speed under the main speed like below. (Could implement the same behaviour as MOM where the prestrafe speed just stays there for the remainder of the map run

![image](https://github.com/surftimer/SurfTimer/assets/9894426/50646789-4dda-4e1e-baa0-1a54b26edc59)
![image](https://github.com/surftimer/SurfTimer/assets/9894426/4aa4bb62-8f1c-40ca-b5ca-e3dbabd1e747)

**Describe alternatives you've considered**

- [x] Getting access to Momentum Mod
- [x] Just dealing with the eye movement
- [x] Moving the chat window to a more central location
- [x] Learning game development, the source engine and C

I've searched for everything that I can think of relating to this but haven't found anything, I'm all open to suggestions/ideas if there are any alternatives!

**Additional context**

I'm mainly curious - _is this possible at all_? I'm assuming there was a good reason why the prestrafe is going to the chat in the first place.   
I reckon this is a potentially huge win for low-medium level surf players as they learn to prestrafe, would love this to be implemented!

I'm happy to give this a go in my free time if it's possible/there are any pointers for how it might be accomplished?

tslashd:
It should be possible, try this https://github.com/surftimer/SurfTimer/pull/593

I don't think there is a particular reason why it is shown only in chat.
As for it showing until the end of the map/bonus run, I personally do not think it is a good idea as the size of the HUD text is not configurable iirc 😞

tmck-code:
Thanks so much for your reply! Will test this out

PR #593, opened 2023-08-01
[Addition] Show Prestrafe speed in CSD

tslashd:
As per #592 .
Disappears after `2` seconds, works only when leaving **Map** or **Bonus** start zones 

image

### DIFF
--- a/addons/sourcemod/scripting/SurfTimer.sp
+++ b/addons/sourcemod/scripting/SurfTimer.sp
@@ -581,6 +581,12 @@ public void OnClientDisconnect(int client)
 		return;
 	}
 
+	// Clear hudPrestrafe value
+	if (strlen(g_szPrespeedValue[client]) != 0)
+	{
+		FormatEx(g_szPrespeedValue[client], sizeof(g_szPrespeedValue), "");
+	}
+
 	// Database
 	if (IsValidClient(client) && !g_bRenaming)
 	{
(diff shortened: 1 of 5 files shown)


addons/sourcemod/scripting/SurfTimer.sp
```diff
@@ -581,6 +581,12 @@ public void OnClientDisconnect(int client)
 		return;
 	}
 
+	// Clear hudPrestrafe value
+	if (strlen(g_szPrespeedValue[client]) != 0)
+	{
+		FormatEx(g_szPrespeedValue[client], sizeof(g_szPrespeedValue), "");
+	}
+
```
Bara: Can be replaced with `g_szPrespeedValue[client][0] = '\0';`

addons/sourcemod/scripting/surftimer/timer.sp
```diff
@@ -654,3 +654,14 @@ public Action DatabaseUpgrading(Handle timer)
 
 	return Plugin_Handled;
 }
+
+// Clear global prestrafe var for the client
+public Action hudPrestrafe(Handle timer, any client)
+{
+	if(IsValidClient(client) && IsClientInGame(client))
+	{
+		FormatEx(g_szPrespeedValue[client], sizeof(g_szPrespeedValue), "");
```
Bara: Same here with `g_szPrespeedValue[client][0] = '\0';`

Bara (approved):

